PR #20, opened 2018-03-22
Sample YAML for running node-disk-manager as k8s DaemonSet.

### LINKED ISSUE #7: Implement node-disk-manager DaemonSet
node-disk-manager should be run as DaemonSet within a Kubernetes Cluster. The user should be able to specify the nodes on which the node-disk-manager is required to run - potentially dedicating some nodes to storage.

The task is to produce an K8s YAML file.

a4abhishek:
Fixes #7 

Made the following assumptions:
- A docker image will be available with latest tag at openebs/node-disk-manager:latest
- The ndm binary will be included in the docker image at /usr/sbin. 
- `ndm start` will start the node-disk-manager service. 
- Kubernetes Clusters will have the MountPropagation feature enabled

Tested in minikube using the following steps:
- Made docker image with `make`.
- Ran minikube with this command: `sudo minikube start --vm-driver=none --feature-gates=MountPropagation=true`
- In YAML changed value of `image` to match with newly built image and `imagePullPolicy` to `IfNotPresent`.
- Applied the YAML
- Checked pod description and log.
- Ran `lsblk` and `ndm` in both inside and outside the pod and match the results.
- Attached Pen Drive, then see if it is visible on the host and inside the pod
- Did mount and unmount operation from within pod and check that it is reflected on the host. 

Signed-off-by: Abhishek Kashyap <abhishek.kashyap@mayadata.io>

### DIFF
--- a/node-disk-manager.yaml
+++ b/node-disk-manager.yaml
@@ -1,0 +1,43 @@
+apiVersion: extensions/v1beta1
+kind: DaemonSet
+metadata:
+  name: node-disk-manager
+spec:
+  template:
+    metadata:
+      labels:
+        app: node-disk-manager
+    spec:
+      # By default the node-disk-manager will be run on all kubernetes nodes
+      # If you would like to limit this to only some nodes, say the nodes 
+      # that have storage attached, you could label those node and use nodeSelector.
+      # Example: Label the storage nodes with - "openebs.io/nodegroup"="storage-node"
+      # kubectl label node <node-name> "openebs.io/nodegroup"="storage-node"
+      #nodeSelector:
+      #  "openebs.io/nodegroup": "storage-node"
+      containers:
+      - name: node-disk-manager
+        command:
+        - /usr/sbin/ndm
+        - start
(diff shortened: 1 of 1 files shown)


a4abhishek:
Addressing issue: https://github.com/openebs/node-disk-manager/issues/7

gkGaneshR:
Can you please provide the link of Dockerfile?

node-disk-manager.yaml
```diff
@@ -0,0 +1,44 @@
...
+        hostPath:
+          path: /dev
+          type: Directory
+      - name: mount
+        hostPath:
+          path: /mnt
+          type: Directory
+      - name: etc
+        hostPath:
+          path: /etc
```
gkGaneshR: Please experiment by sharing mtab or fstab files and check if it works, instead of sharing whole /etc

a4abhishek: @gkGaneshR I did what you suggested. Thanks a lot for your help.

PR #22, opened 2018-03-23
Add a dockerfile to build node-disk-manager

sagarkrsd:
- This PR will add a dockerfile to build node-disk-manager
- Binary is copied from bin/amd64/ndm to /usr/sbin/ndm
- Port 9090 is exposed

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,15 @@
+#
+# This Dockerfile builds node-disk-manager
+# 
+
+#Set the base image
+FROM ubuntu:16.04
+
+#Copy binary to /usr/sbin/ndm
+COPY bin/amd64/ndm /usr/sbin/ndm
+
+#Expose port 9090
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,15 @@
+#
+# This Dockerfile builds node-disk-manager
+# 
+
+#Set the base image
+FROM ubuntu:16.04
```
kmova: Can you make this to be configurable like:
- Use a custom OS image with some pre-installed libraries:
  Refer: https://github.com/kubernetes/node-problem-detector/blob/master/Dockerfile.in

- Make the ARCH (amd64) configurable:
  Refer: https://github.com/kubernetes-incubator/spartakus/blob/master/Dockerfile

sagarkrsd: sure @kmova

PR #37, opened 2018-04-10
unify disk resource name.

pawanpraka1:
NDM is having separate logic to create UUID for the disk resources
which does not have the Wwn number. This patch will unify them
and have same kind of name for all kinds of disk resources.

```
$  kubectl get disks
NAME                                    AGE
disk-60118c2e5b9fe3c33884765ea95db585   7m
disk-dd038c382a92a22f75910b4198d18db2   3s
```

Earlier it was
```
$ kubectl get disk
NAME                                    AGE
disk-0x500003949d304e06                 45s
disk-b3f828db7cd829ac237e90ccf433bd17   5s
```

### DIFF
--- a/cmd/controller/ctrl.go
+++ b/cmd/controller/ctrl.go
@@ -59,13 +59,7 @@ func NewController(
 
 // getUid will return unique id for the disk block device
 func getUid(blk v1.Blockdevice) string {
-	uuid := NDMPrefix
-	if blk.Wwn != "" {
-		uuid += blk.Wwn
-	} else {
-		uuid += util.Hash(blk.Serial)
-	}
-	return uuid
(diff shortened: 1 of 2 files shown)


cmd/controller/ctrl.go
```diff
@@ -59,13 +59,7 @@ func NewController(
 
 // getUid will return unique id for the disk block device
 func getUid(blk v1.Blockdevice) string {
```
ShovanMaity: Make it as exported function. This will use for disk label in iostat metrics.

pawanpraka1: lets keep that change as a part of iostat metrics commit only, as this function is not needed to be exported here in this change.

ShovanMaity: Ok.

Issue #33, opened 2018-04-04
node-disk-manager make file delete all binary at $GOPATH/bin for make in local machine.


shovanmaity:
```
clean: header
@echo '--> Cleaning directory...' 
rm -rf bin
rm -rf ${GOPATH}/bin/${NODE_DISK_MANAGER}
rm -rf ${GOPATH}/pkg/*
@echo '--> Done cleaning.'
```
This code will delete all binary present in go path as NODE_DISK_MANAGER not exported as env variable(For local setup.).

Issue #36, opened 2018-04-09
remove dependency on lsblk command to get the disk related info.


pawanpraka1:
The current implementation of node-disk-manager depends on the lsblk command to get the disk related info. It takes those info from the lsblk command and pushes them to the etcd. We should identify those info directly from the system, instead of forking and execing the lsblk command and parsing the output.

One way to do this is, read the /sys/block/ entries and get the details. Sample code to get if disk is rotational or not 
```
        blockdevices, err := ioutil.ReadDir("/sys/block/")
	if err == nil {
		for _, bdev := range blockdevices {
			fname := "/sys/block/" + bdev.Name() + "/queue/rotational"
			bytes, err := ioutil.ReadFile(fname)
			if err != nil {
				// error
			}
			if bytes[0] == byte('0') {
				// Non-rotational storage is present
			}
		}
       }
```

PR #48, opened 2018-05-22
remove dependency on lsblk

pawanpraka1:
here using libudev to get disk attributes instead of forking/execing lsblk command, which gives us more control over what attributes we want as we can use other libraries like smart attributes to get the other disk properties. 
Not all the attributes are available inside the container, so mounting /run/udev in deployment yaml (node-disk-manager.yaml). Also ndm list CLI is changed to print the data which is stored in etcd.

$ ndm device list
Path: /dev/sdb
Size: 15597568000
Status: Inactive
Model: Ultra
Serial: SanDisk.Ultra...........4C531001520618104171
Vendor: SanDisk

Path: /dev/sda
Size: 500107862016
Status: Active
Model: TOSHIBA_MK5061GS
Serial: 23FCYE2IF
Vendor: ATA


cmd/controller/ctrl.go
```diff
@@ -163,11 +188,12 @@ func deleteDR(name string, c *Controller) {
 }
 
 // deactivateStaleDiskResource deactivates the stale entry from etcd
-func deactivateStaleDiskResource(c *Controller, detected v1.BlockDeviceInfo, listDR *apis.DiskList) {
+func deactivateStaleDiskResource(c *Controller, devices []*udev.Udevice, listDR *apis.DiskList) {
 	for _, item := range listDR.Items {
 		var uuid string
-		for _, blk := range detected.Blockdevices {
-			uuid = getUid(blk)
+		for i := range devices {
```
ShovanMaity: You can use `for _ ,device := range devices {`
instead of
`for i := range devices {
		device := devices[i]`

cmd/controller/ctrl.go
```diff
@@ -190,17 +216,28 @@ func getExistingResource(uuid string, listDR *apis.DiskList) *apis.Disk {
...
 // addNewDiskResource add the newly identified disks to the etcd
-func addNewDiskResource(c *Controller, detected v1.BlockDeviceInfo, listDR *apis.DiskList) {
-	for _, blk := range detected.Blockdevices {
-		if blk.Type == "disk" {
-			uuid := getUid(blk)
-			n, err := strconv.ParseInt(blk.Size, 10, 64)
+func addNewDiskResource(c *Controller, devices []*udev.Udevice, listDR *apis.DiskList) {
+	for i := range devices {
+		device := devices[i]
+		if device.Devtype() == "disk" && device.PropertyValue("SCSI_TYPE") == "disk" {
```
ShovanMaity: `disk` can be used as a constant string
also all the keys which are used to get disk info

pkg/udev/udev.go
```diff
@@ -0,0 +1,138 @@
...
+	// If passed a NULL pointer, return nil
+	if ptr == nil {
+		return nil
+	}
+	// Create a new device object
+	d = &Udevice{
+		ptr: ptr,
+	}
+	// Return the device object
+	return
```
avishnu: Shouldn't the object be returned here?

avishnu: Got it. You may ignore the comment.

shovanmaity:
fixes https://github.com/openebs/node-disk-manager/issues/33

shovanmaity:
fixes - https://github.com/openebs/node-disk-manager/issues/36

PR #54, opened 2018-06-21
add codecov

### LINKED ISSUE #35: add code coverage report for node-disk-manager
While running the travis build we are doing some UT and e2e testing. We want to see the code coverage report to know how much part of it we are covering as a part of out testing framework.

shovanmaity:
Fix #35 

Signed-off-by: Shovan <shovan.cse91@gmail.com>

### DIFF
--- a/hack/test.sh
+++ b/hack/test.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+#
+# This script runs tests and generates a report file.
+set -e
+echo "" > coverage.txt
+PACKAGES=$(go list ./... | grep -v '/vendor/\|/pkg/apis/\|/pkg/client/\|integration_test')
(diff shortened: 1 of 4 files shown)


hack/test.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+#
+# This script runs tests and generates a report file.
+set -e
+echo "" > coverage.txt
+for d in $(go list ./... | grep -v '/vendor/' | grep -v 'integration_test'); do
```
sagarkrsd: I think it would be better to assign the listing of go files to a new variable and then use it in the for loop .

shovanmaity: done.

pawanpraka1:
looks good.

PR #61, opened 2018-07-16
Filter os disk while creating/updating disk cr.

shovanmaity:
This PR contains plugable filter concept and os disk filter.
Filter interface is implemented in this PR.
In `Start()` of each filter we set range of check variables for that filter.
In `Include(*DiskInfo) bool` if property of disk matches with include value it returns true.
In `Exclude(*DiskInfo) bool` if property of disk does not match with exclude value it returns true.
If any of `Include()` or `Exclude()` returns true filter returns true.

**special notes for reviewer** -
I removed package level init for registering probes and filters. If I do like that then we will lost control of registering order as all of them are as go routine. And we want to register all filters then all probes so I have made separate register method of each probe and filters.

**Implementation steps** -
1. read `/proc/self/mounts` and find which partition is mounted on `/` path `ie - sda2`
2. read link `/sys/class/block/[partition-name]` ie - ```/sys/class/block/sda2```
this link looks like - 
```/sys/devices/pci0000:00/0000:00:1f.2/ata1/host0/target0:0:0/0:0:0:0/block/sda/sda2```
read `/` separated entry after ```block``` that is os disk name
3. we need to mount `/proc/self/mounts` inside container if we use self it points to currently running process so content of that file getting changed for that. We need to `mount /proc/1/mounts` this is the proc file of process 1 of host/node.
4. If process 3 is not able to get OS Disk uuid then it search for ```/etc/hosts``` mount point of containers ```/proc/self/mounts``` file. suppose ```/dev/sdc1``` on host is ```/var```, which is where ```/var/lib/docker``` resides and where Docker keeps certain data, such as the hosts file allocated to container. The hosts file is exposed as a bind mount inside the container at ```/etc/hosts``` mount point:
5. one new attribute IgnoreDiskUuid added in Controller struct at the time of event processing if the uuid of the disk is present in IgnoreDiskUuid then that event will not processed by event handler.
6. as a part of creating controller struct all ignored disk's uuid added in IgnoreDiskUuid field of Controller struct and cr creation of these disk is ignored.

For details - [Design doc](https://docs.google.com/document/d/1uyul5Sb6LL5PfXBE7tOUa4jSVLk-aPutZuP8kGRjGX0)

### DIFF
--- a/cmd/controller/filter.go
+++ b/cmd/controller/filter.go
@@ -1,0 +1,94 @@
+/*
+Copyright 2018 OpenEBS Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"github.com/golang/glog"
+)
+
+// Filter contains name, state and filterInterface
+type Filter struct {
+	Name      string          // Name is the name of the filter
+	State     bool            // State is the State of the filter
+	Interface FilterInterface // Interface contains registerd filter
+}
+
+/*
+	ApplyFilter returns true if both any of include() or exclude() returns true.
+	We are having two types of filter function one is inclusion and exclusion type
+	if any of them returns true then filter doesn't want further process of that event.
+*/
+func (f *Filter) ApplyFilter(diskInfo *DiskInfo) bool {
+	return f.Interface.Include(diskInfo) || f.Interface.Exclude(diskInfo)
+
+}
+
+// Start implements FilterInterface's Start()
+func (f *Filter) Start() {
+	f.Interface.Start()
+}
+
+// FilterInterface contains Filters interface and Start()
+type FilterInterface interface {
+	Start()
+	Filters
+}
+
+/*
+	Filters contains Include() and Exclude() filter method. There
+	will be some preset value of include and exclude if passing DiskInfo
+	matches with include value then it returns true if passing DiskINfo
+	does not match with exclude value then it returns false
+*/
+type Filters interface {
+	Include(*DiskInfo) bool // Include returns True if passing DiskInfo matches with include value
+	Exclude(*DiskInfo) bool // exclude returns True if passing DiskInfo does not matche with exclude value
+}
+
+// AddNewFilter adds new filter to controller object
+func (c *Controller) AddNewFilter(filter *Filter) {
+	c.Lock()
+	defer c.Unlock()
+	filters := c.Filters
+	filters = append(filters, filter)
+	c.Filters = filters
+	glog.Info(filter.Name, " applied")
+}
+
+// ListFilter returns list of active filters associated with controller object
+func (c *Controller) ListFilter() []*Filter {
+	c.Lock()
+	defer c.Unlock()
+	listFilter := make([]*Filter, 0)
+	for _, filter := range c.Filters {
+		if filter.State {
+			listFilter = append(listFilter, filter)
+		}
+	}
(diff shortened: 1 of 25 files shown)

### CODE REVIEW

File: cmd/controller/osdisk.go [hunk 1]
```diff
@@ -0,0 +1,104 @@
+/*
```
shovanmaity: @kmova This file contains functions to get uuid using mounts file and mount point. This is not related to controller. Thinking where to put this file inside util or some other package ... your suggestion

cmd/controller/filter.go
```diff
@@ -0,0 +1,78 @@
...
+		if filter.FilterState {
+			listFilter = append(listFilter, filter)
+		}
+	}
+	return listFilter
+}
+
+// ShouldProcess checks status for every registered filters if all filters allow
+// for their filter condition it returns true else it returns false
+func (c *Controller) ShouldProcess(diskDetails *DiskInfo) bool {
```
shovanmaity: ApplyFilter

cmd/controller/probe.go
```diff
@@ -94,3 +94,13 @@ func (c *Controller) ListProbe() []*Probe {
 	}
 	return listProbe
 }
+
+// FillDetails lists registered probes and fills details from each probe
+func (c *Controller) FillDetails(diskDetails *DiskInfo) {
```
shovanmaity: FillDeviceDetails

cmd/filter/filter.go
```diff
@@ -0,0 +1,59 @@
...
+
+const (
+	defaultEnabled  = true  // use in each filter to make it enable.
+	defaultDisabled = false // use in each filter to make it disable.
+)
+
+// RegisterdFilters contains register function of filters which we want to register
+var RegisterdFilters = []func(){osDiskFilterRegister}
+
+type registerFilter struct {
```
shovanmaity: remove prefix

cmd/controller/filter.go
```diff
@@ -0,0 +1,94 @@
...
+
+// Start implements FilterInterface's Start()
+func (f *Filter) Start() {
+	f.Interface.Start()
+}
+
+// FilterInterface contains ApplyFilter interface and Start()
+type FilterInterface interface {
+	Start()
+	ApplyFilter
```
kmova: The struct ApplyFilter is clashing with the function ApplyFilter. Maybe the struct can be renamed to "Filters" or even totally eliminate it. bring up the Include and Exclude functions into the FilterInterace itself.

cmd/filter/osdiskfilter.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// Include checks if disk uuid of passing DiskInfo struct is same as
+// uuid mention in osDiskFilter struct if it matches it returns true
+// else it returns false
+func (odf *osDiskFilter) Include(d *controller.DiskInfo) bool {
+	return odf.devPath == d.Path
+}
+
+// Include checks if disk uuid of passing DiskInfo struct is same as
```
kmova: Fix these comments as they apply to OSdisk

cmd/probe/eventhandler.go
```diff
@@ -43,21 +43,13 @@ func (pe *ProbeEvent) addDiskEvent(msg controller.EventMessage) {
...
-		for _, probe := range probes {
-			glog.Info("disk details filled by ", probe.ProbeName)
-			probe.FillDiskDetails(diskDetails)
-		}
-		diskApi := diskDetails.ToDisk()
-		oldDr := pe.Controller.GetExistingResource(diskList, diskDetails.ProbeIdentifiers.Uuid)
-		if oldDr != nil {
-			pe.Controller.UpdateDisk(diskApi, oldDr)
+		pe.Controller.FillDiskDetails(diskDetails)
+		if pe.Controller.ApplyFilter(diskDetails) {
```
kmova: Before this call, add a comment on what is being one in apply filter

cmd/controller/filter.go
shovanmaity: done.

cmd/filter/osdiskfilter.go
shovanmaity: done

cmd/probe/eventhandler.go
shovanmaity: done

PR #62, opened 2018-07-17
Refactoring integration test

a4abhishek:
Update integration test logic to align with CITF

This PR is also intended to make Integration Test of NDM compatible with multiple pods and multiple containers.

- **new file:   integration_test/common_resource/common_resource.go**
This package will hold all the common variables and constants so that all other packages can use them. Since this package is intended to be used by all other package, we should keep this on the top of the package hierarchy to avoid cyclic dependency.

- **modified:   integration_test/ginkgo_integration_test.go**
**modified:   integration_test/k8s_util/ndm_specific.go**
**modified:   integration_test/ndm_util/ndm_util.go**
These files are modified to use `CITF` for specific operations and common_resource package for common thing e.g. `CitfInstance`.

- **deleted:    integration_test/common/k8s_specific.go**
**deleted:    integration_test/common/string_operations.go**
**deleted:    integration_test/common/system_commands.go**
**deleted:    integration_test/k8s_util/k8s_util.go**
**deleted:    integration_test/minikube_adm/minikube_adm.go**
These files are no longer used because the functionality these files used to serve are now served by CITF in even better manner.


sagarkrsd:
@a4abhishek I think there are some files which are a candidate for refactoring having 60 or more lines of code.
I think these should be resolved, may be it can cause problems for coming PRs.

cc @kmova

a4abhishek:
@sagarkrsd please check if current status of bettercodehub is OK. Please feel free to comment if its not.

PR #68, opened 2018-08-16
Include support for creating sparse file

kmova:
For testing with nodes that have no additional disks, sparse disks provide a
mechanism to create a sparse file and use as a disk resource.

The NDM sparse file creation can be configured via the following
environment variables:
- SPARSE_FILE_DIR : A valid directory where sparse files need to be created
- SPARSE_FILE_SIZE : Size of each sparse file in byte. Defaults to 1GB
- SPRASE_FILE_COUNT: Number of files to be created. Defaults to 1

To differentiate between sparse disks and regular disk, a disk-type label
as been introduced. For sparse disks, this disk-type will be `sparse`, where
as for regular disks it will be set as `disk`.

Like disk resources, the sparse disk resource's state will be marked as
`Uknown` when NDM is shutdown and is made `Active` when the NDM is restarted
on the same node.

Note:
- If a sparse disk already exists in the directory, it will be reused as is.
- The naming convention for sparse disk is <user specified dir>/<count>-ndm-sparse.img
  count starts with 0 to n-1, where n is the count specified by user as
  SPARSE_FILE_COUNT

The Sparse Disk CR is as follows:
```
- apiVersion: openebs.io/v1alpha1
  kind: Disk
  metadata:
    clusterName: ""
    creationTimestamp: 2018-08-16T14:54:06Z
    labels:
      kubernetes.io/hostname: gke-kmova-helm-default-pool-5116437a-l3rp
      ndm.io/disk-type: sparse
    name: sparsef594eb00eecf63247f655aaf524bd46b
    namespace: ""
    resourceVersion: "34776"
    selfLink: /apis/openebs.io/v1alpha1/sparsef594eb00eecf63247f655aaf524bd46b
    uid: 393cfa91-a164-11e8-b06f-42010a80004b
  spec:
    capacity:
      logicalSectorSize: 0
      storage: 1073741824
    details:
      firmwareRevision: ""
      model: ""
      serial: ""
      spcVersion: ""
      vendor: ""
    path: /var/openebs/0-ndm-sparse.img
  status:
    state: Active
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```

### DIFF
--- a/cmd/controller/sparsefilegenerator.go
+++ b/cmd/controller/sparsefilegenerator.go
@@ -1,0 +1,180 @@
+/*
+Copyright 2018 OpenEBS Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"github.com/golang/glog"
+	"github.com/openebs/node-disk-manager/pkg/util"
+
+	"fmt"
+	"os"
+	"path"
+	"strconv"
+)
+
+/*
+Sparse File help simulate disk objects that can be used for testing and
+proto typing solutions built using node-disk-manager(NDM).  Sparse files
+will be created if NDM is provided with the location where sparse files
+should be located.
+
+On Startup, if a sparse directory (SPARSE_FILE_DIR) is specified as a
+environment variable, a Sparse file with specified size (SPARSE_FILE_SIZE) will
+be created and an associated Disk CR will be added to Kubernetes. By default
+only one sparse file will be created which can be changed by passing the desired
+number of sparse files required via the environment variable SPARSE_FILE_COUNT.
+
+On Shutdown, the status of the sparse file Disk CR will be marked as Unknown.
+*/
+
+const (
+	ENV_SPARSE_FILE_DIR   = "SPARSE_FILE_DIR"
+	ENV_SPARSE_FILE_SIZE  = "SPARSE_FILE_SIZE"
+	ENV_SPARSE_FILE_COUNT = "SPARSE_FILE_COUNT"
+
+	SPARSE_FILE_NAME          = "ndm-sparse.img"
+	SPARSE_FILE_DEFAULT_SIZE  = "1073741824"
+	SPARSE_FILE_DEFAULT_COUNT = "1"
+
+	SPARSE_DISKTYPE    = "sparse"
+	SPARSE_DISK_PREFIX = "sparse-"
+)
+
+// GetSparseFileDir returns the full path to the sparse
+//  file directory on the node.
+func GetSparseFileDir() string {
+
+	sparseFileDir := os.Getenv(ENV_SPARSE_FILE_DIR)
+
+	if len(sparseFileDir) < 1 {
+		return ""
+	}
+
+	info, err := os.Stat(sparseFileDir)
+	if os.IsNotExist(err) || !info.Mode().IsDir() {
+		glog.Info("Specified directory doesnt exist:  ", sparseFileDir)
+		return ""
+	}
+
+	return sparseFileDir
+}
+
+// GetSparseFileCount returns the number of sparse files to be
+//  created by NDM. Returns 0, if invalid count is specified.
+func GetSparseFileCount() int {
+
+	sparseFileCountStr := os.Getenv(ENV_SPARSE_FILE_COUNT)
+
+	if len(sparseFileCountStr) < 1 {
+		sparseFileCountStr = SPARSE_FILE_DEFAULT_COUNT
+	}
+
+	sparseFileCount, econv := strconv.Atoi(sparseFileCountStr)
+	if econv != nil {
+		glog.Info("Error converting sparse file count:  ", sparseFileCountStr)
+		return 0
+	}
+
+	return sparseFileCount
+}
+
+// GetSparseFileSize returns the size of the sparse file to be
+//  created by NDM. Returns 0, if invalid size is specified.
+func GetSparseFileSize() int64 {
+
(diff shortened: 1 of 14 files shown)


cmd/controller/sparsefilegenerator.go
```diff
@@ -0,0 +1,180 @@
...
+	} else {
+		glog.Info("Sparse file already exists: ", sparseFileInfo.Name())
+	}
+	return err
+}
+
+// GetSparseDiskUuid returns a fixed UUID for the sparse
+//  disk on a given node.
+func GetSparseDiskUuid(hostname string, sparseFile string) string {
+	return SPARSE_DISK_PREFIX + util.Hash(hostname+sparseFile)
```
shovanmaity: why we are changing prefix as we are adding disk type label is it not a duplicate entry or is it for visual only.

kmova: For ease of use only @shovanmaity

cmd/controller/sparsefilegenerator.go
```diff
@@ -0,0 +1,180 @@
...
+)
+
+// GetSparseFileDir returns the full path to the sparse
+//  file directory on the node.
+func GetSparseFileDir() string {
+
+	sparseFileDir := os.Getenv(ENV_SPARSE_FILE_DIR)
+
+	if len(sparseFileDir) < 1 {
+		return ""
```
avishnu: Will it be a good idea to have a default sparse dir as well (on similar lines as default count and default size) ? Also, create if not exists.

avishnu: Or is the idea to use sparse_file_dir as the decision flag to go ahead with sparse-file creation? If this is the idea, then its fine and you can ignore above comment.

cmd/controller/sparsefilegenerator.go
```diff
@@ -0,0 +1,180 @@
...
+	sparseFileSizeStr := os.Getenv(ENV_SPARSE_FILE_SIZE)
+	if len(sparseFileSizeStr) < 1 {
+		sparseFileSizeStr = SPARSE_FILE_DEFAULT_SIZE
+	}
+
+	sparseFileSize, econv := strconv.ParseInt(sparseFileSizeStr, 10, 64)
+	if econv != nil {
+		glog.Info("Error converting sparse file size:  ", sparseFileSizeStr)
+		return 0
+	}
```
avishnu: Should it bail out if the size is less than a expected minimum value ?

cmd/controller/sparsefilegenerator.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+// CheckAndCreateSparseFile will reuse the existing sparse file if it already exists,
+//   for handling cases where NDM is upgraded or restarted. If the file doesn't exist
+//   a new file will be created.
+func CheckAndCreateSparseFile(sparseFile string, sparseFileSize int64) error {
+	sparseFileInfo, err := util.SparseFileInfo(sparseFile)
+	if err != nil {
+		err = util.SparseFileCreate(sparseFile, sparseFileSize)
```
avishnu: Can we have a log here along-with the error? Also mention in the log "proceeding with sparse-file creation..."

cmd/controller/sparsefilegenerator.go
```diff
@@ -0,0 +1,180 @@
...
+		c.MarkSparseDiskStateActive(sparseFile, sparseFileSize)
+	}
+}
+
+// CheckAndCreateSparseFile will reuse the existing sparse file if it already exists,
+//   for handling cases where NDM is upgraded or restarted. If the file doesn't exist
+//   a new file will be created.
+func CheckAndCreateSparseFile(sparseFile string, sparseFileSize int64) error {
+	sparseFileInfo, err := util.SparseFileInfo(sparseFile)
+	if err != nil {
```
avishnu: Can we have a log here along-with the error? Also mention in the log "proceeding with sparse-file creation..."

cmd/controller/sparsefilegenerator.go
```diff
@@ -0,0 +1,180 @@
...
+		glog.Info("Error fetching the size of sparse file: ", err)
+		glog.Error("Failed to create a disk CR for sparse file: ", sparseFile)
+		return
+	}
+
+	diskDetails.Capacity = uint64(sparseFileInfo.Size())
+
+	//If a Disk CR already exits, update it. If not create a new one.
+	c.CreateDisk(diskDetails.ToDisk())
```
avishnu: Does this call always return success ?

cmd/controller/sparsefilegenerator.go
```diff
@@ -0,0 +1,180 @@
...
+
+	glog.Info("Created Disk CR for Sprase Disk: ", diskDetails.Uuid)
```
avishnu: Typo - Sparse

cmd/controller/sparsefilegenerator.go
kmova: Yes, the idea is to use this also as a flag to create sparse files.

cmd/controller/sparsefilegenerator.go
kmova: Could do this @avishnu -- can we assume 1G as minimum or probably go with user defined value for now and add the minimum value check based on usage.

cmd/controller/sparsefilegenerator.go
kmova: Good one. Added.

cmd/controller/sparsefilegenerator.go
kmova: Using an existing method here. If there is an error creating the CR, it logs an error.

cmd/controller/sparsefilegenerator.go
kmova: cool! fixed.

cmd/controller/sparsefilegenerator.go
avishnu: The log in the next line may become ambiguous if the CreateDisk fails.

cmd/controller/sparsefilegenerator.go
vishnuitta: yes.. there are some minimums.. can't be less than 64MB.. so, lets make the minimum as 512MB.. lets take it as 1G as 1G is not too much to ask

kmova: ok. I will go with setting the minimum as 1G.

PR #73, opened 2018-08-31
feat(filters): include/exclude disks based on device path

kmova:
Depending on where NDM is running - cloud or virtual hardware,
the host (udev) modules don't always provide all the required
details.

For instance, GKE nodes display loop back devices as disks,
where as disks in AWS don't always show typical device attributes
like - model, vendor or serial number.

This PR introduces a feature where, user can specify a list of
include/exclude keywords that can appear in the device path. For
exampke, the default configuration provided, excludes loop back
devices from being added as disks.

### DIFF
--- a/cmd/filter/pathfilter.go
+++ b/cmd/filter/pathfilter.go
@@ -31,7 +31,7 @@ var (
 	pathFilterName  = "path filter"  // filter device paths
 	pathFilterState = defaultEnabled // filter state
 	includePaths    = ""
-	excludePaths    = ""
+	excludePaths    = "loop"
 )
 
 // pathFilterRegister contains registration process of PathFilter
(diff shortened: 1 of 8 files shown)


cmd/filter/pathfilter.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+const (
+	pathFilterKey = "path-filter"
+)
+
+var (
+	pathFilterName  = "path filter"  // filter device paths
+	pathFilterState = defaultEnabled // filter state
+	includePaths    = ""
```
shovanmaity: Can we add default exclude loop here if ndm is not able to read from config map then there will be a problem.

kmova: in what scenarios will we not have the config map? Having said that.. it is good idea to add loop as a default filter. Will add it.

kmova: added loop!

Issue #75, opened 2018-09-04
ndm set sparse file related cr's state as inactive if any mismatch occur.


shovanmaity:
ndm makes the state of sparse file related cr as inactive if any mismatch occur. One case is like this if user is using openEBS volumes then wrapper iscsi disk is created and vendor of that disk is openEBS which we are filtering out. When user delete that volume one disk detach event comes and as that disk is not present in etcd it is considered as a mismatch. Then ndm will rescan system and update state of the crs. As rescan is dependent on udev only it will not identify sparse files and make the state of the cr as inactive.

In ndm configmap I added google as a exclude vendor. 2 gpd attached with that vm then I applied ndm-operator.yaml

```
shovan_maity@instance-1:~/demo$ lsblk
NAME   MAJ:MIN RM SIZE RO TYPE MOUNTPOINT
sdb      8:16   0  10G  0 disk 
└─sdb1   8:17   0  10G  0 part 
sdc      8:32   0  10G  0 disk 
sda      8:0    0  10G  0 disk 
└─sda1   8:1    0  10G  0 part /
shovan_maity@instance-1:~/demo$ kubectl describe disk
Name:         sparse-8f6ec02111d05c0aea7c04dc21c4bc88
Namespace:    
Labels:       kubernetes.io/hostname=instance-1
              ndm.io/disk-type=sparse
Annotations:  <none>
API Version:  openebs.io/v1alpha1
Kind:         Disk
Metadata:
  Cluster Name:        
  Creation Timestamp:  2018-09-04T05:07:52Z
  Generation:          1
  Resource Version:    585255
  Self Link:           /apis/openebs.io/v1alpha1/disks/sparse-8f6ec02111d05c0aea7c04dc21c4bc88
  UID:                 7a093772-b000-11e8-a306-42010a8e0003
Spec:
  Capacity:
    Logical Sector Size:  0
    Storage:              1073741824
  Details:
    Firmware Revision:  
    Model:              
    Serial:             
    Spc Version:        
    Vendor:             
  Path:                 /var/openebs/0-ndm-sparse.img
Status:
  State:  Active
Events:   <none>
```

Then I detach one gpd as gpd is filtered out via vendor filter it was a mismatch and then state of the cr was like this- 
```
shovan_maity@instance-1:~/demo$ lsblk
NAME   MAJ:MIN RM SIZE RO TYPE MOUNTPOINT
sdb      8:16   0  10G  0 disk 
[...]

PR #76, opened 2018-09-04
Fix sparse disks state mismatch

shovanmaity:
ndm makes the state of sparse file related cr as inactive if any
mismatch occur. One case is like this if user is using openEBS
volumes then wrapper iscsi disk is created and vendor of that disk
is openEBS which we are filtering out. When user delete that volume
one disk detach event comes and as that disk is not present in etcd
it is considered as a mismatch. Then ndm will rescan system and update
state of the crs. As rescan is dependent on udev only it will not
identify sparse files and make the state of the cr as inactive.

This PR contains the code which will check for sparse files if it is
sparse file related cr and it will make the status of the cr inactive
if sparse file is not present. (fix - #75)


cmd/controller/sparsefilegenerator.go
```diff
@@ -161,6 +164,23 @@ func GetSparseDiskUuid(hostname string, sparseFile string) string {
...
+	sparseFileLocation := GetSparseFileDir()
+	sparseUids := make([]string, 0)
+	files, err := ioutil.ReadDir(sparseFileLocation)
+	if err != nil {
+		glog.Error("Failed to read sperse file names : ", err)
+		return sparseUids
+	}
+	for _, file := range files {
+		if strings.HasSuffix(file.Name(), SPARSE_FILE_NAME) {
+			fileName := path.Join(GetSparseFileDir(), file.Name())
```
a4abhishek: @shovanmaity You don't need to call `GetSparseFileDir` again as you have already did that in the first line of this function.

shovanmaity: done :+1:

cmd/controller/sparsefilegenerator.go
```diff
@@ -161,6 +164,25 @@ func GetSparseDiskUuid(hostname string, sparseFile string) string {
 	return SPARSE_DISK_PREFIX + util.Hash(hostname+sparseFile)
 }
 
+// GetActiveSparseDisksUuids returns UUIDs for the sparse
+// disks present in a given node.
+func GetActiveSparseDisksUuids(hostname string) []string {
+	sparseFileLocation := GetSparseFileDir()
+	sparseUids := make([]string, 0)
```
sagarkrsd: @shovanmaity Can this be sparseUuids , just for naming convention !!

cmd/controller/diskstore.go
```diff
@@ -124,13 +124,14 @@ func (c *Controller) GetExistingResource(listDr *apis.DiskList, uuid string) *ap
...
 func (c *Controller) DeactivateStaleDiskResource(devices []string) {
+	listDisks := append(devices, GetActiveSparseDisksUuids(c.HostName)...)
 	listDR, err := c.ListDiskResource()
 	if err != nil {
 		glog.Error(err)
 		return
 	}
 	for _, item := range listDR.Items {
-		if !util.Contains(devices, item.ObjectMeta.Name) {
+		if !util.Contains(listDisks, item.ObjectMeta.Name) {
```
kmova: Can we test this by checking the disk-type label?

kmova: On second thought, this may be a good idea for now. This logic can catch for deleted sparse files.

cmd/controller/diskstore.go
```diff
@@ -124,13 +124,14 @@ func (c *Controller) GetExistingResource(listDr *apis.DiskList, uuid string) *ap
 // list of active resources. One active resource which is present in etcd not in
 // system that will be marked as inactive.
 func (c *Controller) DeactivateStaleDiskResource(devices []string) {
+	listDisks := append(devices, GetActiveSparseDisksUuids(c.HostName)...)
```
pawanpraka1: this looks OK, but it is supposed to be part of scan function, which is passing devices to the  DeactivateStaleDiskResource. As Scan function's job is to scan every resource that NDM is managing, it could be either block device or sparse file.

PR #78, opened 2018-09-04
add support for virtual disk uuid generation

shovanmaity:
ndm is not able to generate uuid of virtual disks whose model is `Virtual_disk` this pr will add `Virtual_disk` model as a special case and allows ndm to generate uuid of that disk as the way it generates for `EphemeralDisk` model.

Changes committed:
modified:   pkg/udev/common.go

### DIFF
--- a/pkg/udev/common.go
+++ b/pkg/udev/common.go
@@ -107,7 +107,10 @@ func (device *UdevDevice) GetUid() string {
 	// 	On Gke, we have the ID_TYPE property, but still disks will have
 	// same attributes. We have to put a special check to handle it and process
 	// it like a Virtual disk.
-	if len(idtype) == 0 || model == "EphemeralDisk" {
+	localDiskModels := make([]string, 0)
+	localDiskModels = append(localDiskModels, "EphemeralDisk")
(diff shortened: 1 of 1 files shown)


pkg/udev/common.go
```diff
@@ -107,7 +107,10 @@ func (device *UdevDevice) GetUid() string {
 	// 	On Gke, we have the ID_TYPE property, but still disks will have
 	// same attributes. We have to put a special check to handle it and process
 	// it like a Virtual disk.
-	if len(idtype) == 0 || model == "EphemeralDisk" {
+	localDiskModels := make([]string, 0)
+	localDiskModels = append(localDiskModels, "EphemeralDisk")
```
pawanpraka1: @kmova,  now I am afraid if there are more cases like this. If customer is trying with new type of EphemeralDisk disk then we can not help him without making any code changes. Hardcoding it in the code will not scale.
 Can we pass this info from env which we can set in the yaml. We can have VDISK_MODEL_LIST environment, and if it is not set we will use EphemeralDisk and Virtual_disk by default?

PR #79, opened 2018-09-05
fix(startup): wait for disk crd to be loaded before scanning disks

kmova:
As part of the initial startup, if the disk CRDs are not loaded,
the discovery process fails due to errors in updating. A restart
will be required to fix this after the disk CRDs are available in
etcd.

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"os"
 	"sync"
+	"time"
 
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
@@ -42,6 +43,11 @@ const (
 
 const (
 	NDMDefaultDiskType = "disk" // NDMDefaultDiskType will be used to initialize the disk type
+)
+
+const (
+	// CRDRetryInterval is used if CRD is not present.
+	CRDRetryInterval = 10 * time.Second
 )
 
 // ControllerBroadcastChannel is used to send a copy of controller object to each probe.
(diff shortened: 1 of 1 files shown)


cmd/controller/controller.go
```diff
@@ -133,8 +139,26 @@ func (c *Controller) setNodeName() error {
...
+		}
+		glog.Info("Disk CRD is available")
+		break
+	}
+}
+
 // Start is called when we execute cli command ndm start.
 func (c *Controller) Start() {
+	//Wait for Disk CRD to be loaded
+	c.WaitForDiskCRD()
```
pawanpraka1: I think this is not the right place to wait for disk CRD as we are starting the probe (probe.start()) before the controller in the ndm start cli command. Can you verify the logs, you should be seeing "Disk CRD is available" after "disk CRD is pushed to etcd" logs?

kmova: you are right @pawanpraka1 .. I will shift this to the instantiation of controller.

kmova:
Logs after making the above change:

```
kiran_mova@kmova-dev:~/go/src/github.com/kmova/openebs$ kubectl logs -n openebs openebs-ndm-dnzwm | more
E0905 10:05:51.263175       1 controller.go:152] Disk CRD is not available yet. Retrying after 10s, error: the server could not find the requ
ested resource (get disks.openebs.io)
I0905 10:06:01.268492       1 controller.go:156] Disk CRD is available
I0905 10:06:01.268521       1 filter.go:58] registering filters
I0905 10:06:01.268540       1 filter.go:70] configured os disk exclude filter : state enable
I0905 10:06:01.268900       1 filter.go:70] configured vendor filter : state enable
I0905 10:06:01.268919       1 filter.go:70] configured path filter : state enable
I0905 10:06:01.268939       1 probe.go:57] registering probes
I0905 10:06:01.268955       1 probe.go:83] configured smart probe : state enable
I0905 10:06:01.269045       1 probe.go:83] configured udev probe : state enable
I0905 10:06:01.269329       1 udevprobe.go:186] starting udev probe listener
```

PR #80, opened 2018-09-06
make os disk exclude filter configurable via cm

shovanmaity:
make os disk exclude filter configurable via configmap we can specify
mount points using configmap in include section of os disk exclude filter.

Changes committed:
modified:   cmd/filter/osdiskexcludefilter.go
modified:   ndm-operator.yaml


ndm-operator.yaml
```diff
@@ -35,7 +35,8 @@ data:
         {
           "key": "os-disk-exclude-filter",
           "name": "os disk exclude filter",
-          "state": "true"
+          "state": "true",
+          "include": "/,/etc/hosts,/boot"
```
sagarkrsd: Is using include keyword here could be confusing ?

shovanmaity: yes done.

PR #82, opened 2018-09-10
Add logic to show correct SPC version for disks rather than raw value

sagarkrsd:
This PR will introduce the following changes -
-  It will add logic to show the correct SPC version (i.e implemented specifications) of a disk instead of showing the raw version value obtained from scsi inquiry command.

-  It will change field **SPCVersion** to **Compliance** which seems to be a more generic term than SPCVersion.

- It will fix the nil pointer reference error in server_test file which occurs when the port to be checked is already on use and calling net.Listen returns error.

Changes committed:
	
    modified:   cmd/controller/disk.go
	modified:   cmd/probe/smartprobe.go
	modified:   cmd/probe/smartprobe_test.go
	modified:   docs/design.md
	modified:   pkg/apis/openebs.io/v1alpha1/types.go
	modified:   pkg/smart/doc.go
	modified:   pkg/smart/mockdata.go
	modified:   pkg/smart/scsidevice.go
	modified:   pkg/smart/scsiinquirypage.go
	modified:   pkg/smart/types.go
    modified:   pkg/server/server_test.go
	modified:   samples/disk.yaml

### DIFF
--- a/pkg/smart/scsiinquirypage.go
+++ b/pkg/smart/scsiinquirypage.go
@@ -57,7 +57,10 @@ func (d *SCSIDev) scsiInquiry() (InquiryResponse, error) {
 // command using Inquiry response struct
 func (inquiry InquiryResponse) getValue() map[string]string {
 	InqRespMap := make(map[string]string)
-	InqRespMap[SPCVersion] = fmt.Sprintf("%.d", inquiry.Version)
+
+	SPCVersionValue := fmt.Sprintf("%.d", (inquiry.Version - 0x02))
+
+	InqRespMap[Compliance] = "SPC-" + SPCVersionValue
 	InqRespMap[Vendor] = fmt.Sprintf("%.8s", inquiry.VendorID)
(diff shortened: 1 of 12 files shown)


pkg/smart/scsiinquirypage.go
```diff
@@ -57,7 +57,10 @@ func (d *SCSIDev) scsiInquiry() (InquiryResponse, error) {
 // command using Inquiry response struct
 func (inquiry InquiryResponse) getValue() map[string]string {
 	InqRespMap := make(map[string]string)
-	InqRespMap[SPCVersion] = fmt.Sprintf("%.d", inquiry.Version)
+
+	SPCVersionValue := fmt.Sprintf("%.d", (inquiry.Version - 0x02))
```
shovanmaity: we can avoid this extra variable

sagarkrsd: I thought getting the value and concatenating it should be done separately !

shovanmaity: if `SPC-` is constant for all the disks then we can remove extra variable else it is good.

Issue #97, opened 2018-09-30
Fix lint error in sparsefilegenerator.go file


qiell:
[sparsefilegenerator.go](https://github.com/openebs/node-disk-manager/blob/master/cmd/controller/sparsefilegenerator.go)

```
Line 48: warning: exported const ENV_SPARSE_FILE_DIR should have comment (or a comment on this block) or be unexported (golint)
```

Akash4927:
@albertogviana PR is already raised for this

AraanBranco:
@Akash4927 is it still open?

ShaakhaBrahma:
am working on this issue

ShaakhaBrahma:
Signed-off-by: ShaakhaBrahma <vatsavatsi777@protonmail.com>

ShaakhaBrahma:
Signed-off-by: ShaakhaBrahma <vatsavatsi777@protonmail.com>

akhilerm:
Changes are merged as part of multiple PRs

PR #131, opened 2018-10-01
Fix warning

bege13mot:
https://github.com/openebs/node-disk-manager/issues/97

Signed-off-by: bege13mot <bege13mot@gmail.com>

### DIFF
--- a/cmd/controller/sparsefilegenerator.go
+++ b/cmd/controller/sparsefilegenerator.go
@@ -45,6 +45,9 @@ On Shutdown, the status of the sparse file Disk CR will be marked as Unknown.
 */
 
 const (
+	/*ENV_SPARSE_FILE_DIR - if it specified as a environment variable,
+	a Sparse file with specified size (SPARSE_FILE_SIZE) will
+	be created and an associated Disk CR will be added to Kubernetes.*/
 	ENV_SPARSE_FILE_DIR   = "SPARSE_FILE_DIR"
 	ENV_SPARSE_FILE_SIZE  = "SPARSE_FILE_SIZE"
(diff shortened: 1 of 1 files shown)


cmd/controller/sparsefilegenerator.go
```diff
@@ -45,6 +45,9 @@ On Shutdown, the status of the sparse file Disk CR will be marked as Unknown.
 */
 
 const (
+	/*ENV_SPARSE_FILE_DIR - if it specified as a environment variable,
```
shovanmaity: mention that it will create Sparse file inside `ENV_SPARSE_FILE_DIR` dir

bege13mot: Thank you for your review. It is fixed

cmd/controller/sparsefilegenerator.go
```diff
@@ -45,6 +45,10 @@ On Shutdown, the status of the sparse file Disk CR will be marked as Unknown.
...
+	a Sparse file with specified size (SPARSE_FILE_SIZE) will
+	be created inside specified directory (ENV_SPARSE_FILE_DIR)
+	and an associated Disk CR will be added to Kubernetes.*/
```
sagarkrsd: @bege13mot  Providing comment for only one const will not fix the lint error. Please provide comments for all, otherwise  it will still give the error for next const i.e. ENV_SPARSE_FILE_SIZE here.

bege13mot: Hi @sagarkrsd ,
Thank you for the review. It is fixed.

Issue #111, opened 2018-09-30
Fix go vet errors in fdutil_test.go


Akash4927:
[fdutil_test.go](https://github.com/openebs/node-disk-manager/blob/master/pkg/util/fdutil_test.go)

```
Line 31: error: syscall.FdSet composite literal uses unkeyed fields (vet)
Line 32: error: syscall.FdSet composite literal uses unkeyed fields (vet)
Line 33: error: syscall.FdSet composite literal uses unkeyed fields (vet)
Line 50: error: syscall.FdSet composite literal uses unkeyed fields (vet)
Line 51: error: syscall.FdSet composite literal uses unkeyed fields (vet)
Line 52: error: syscall.FdSet composite literal uses unkeyed fields (vet)
Line 53: error: syscall.FdSet composite literal uses unkeyed fields (vet)
Line 54: error: syscall.FdSet composite literal uses unkeyed fields (vet)
Line 69: error: syscall.FdSet composite literal uses unkeyed fields (vet)
Line 70: error: syscall.FdSet composite literal uses unkeyed fields (vet)
Line 71: error: syscall.FdSet composite literal uses unkeyed fields (vet)
```

PR #134, opened 2018-10-01
style(ndm): Fix go vet errors in fdutil_test.go

bege13mot:
https://github.com/openebs/node-disk-manager/issues/111

Signed-off-by: bege13mot <bege13mot@gmail.com>

### DIFF
--- a/pkg/util/fdutil_test.go
+++ b/pkg/util/fdutil_test.go
@@ -23,18 +23,26 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+func fdGetArray() []syscall.FdSet {
+	results := make([]syscall.FdSet, 0)
+	fd1, fd2, fd6, fd128 := syscall.FdSet{}, syscall.FdSet{}, syscall.FdSet{}, syscall.FdSet{}
+	fd1.Bits[0] = 1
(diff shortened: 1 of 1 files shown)


shovanmaity:
@bege13mot thanks for this pr, changes looks good. You can change function `getFdArray` so that it can take value and return one fdset for that value it will be more generic.
something like -
```
func getFdSet(value int64) syscall.FdSet {
	// return fdset for value
}
```

shovanmaity:
Yes It will be good if you make those changes. thanks again :+1: . Just thinking why it requires `int32`
It should take `int64` - https://golang.org/src/syscall/ztypes_linux_amd64.go#L546 can you put some error log or message

bege13mot:
Hi @shovanmaity ,

As I understand, `syscall.FdSet` depends on the system platform. And on my machine it `int32`. Like https://golang.org/src/syscall/ztypes_darwin_amd64.go

shovanmaity:
> Hi @shovanmaity ,
> 
> As I understand, `syscall.FdSet` depends on the system platform. And on my machine it `int32`. Like https://golang.org/src/syscall/ztypes_darwin_amd64.go

@bege13mot changes are good as it is dependent on system no need to update. +1

PR #149, opened 2018-10-17
fix(smartProbe):fix smart probe to fill larger(>2TB) disk capacity also

sagarkrsd:
smart probe is now able to fill disk capacity for disks having capacity larger than 2 TB also.

Earlier it was not able to fill disk capacity of a disk having capacity more than 2TB since it was only using ReadCapacity10 scsi command to get the capacity of a disk.
Now it is able to send ReadCapacity16 scsi command also if  the capacity is more than 2TB which is the maximum limit fetched using ReadCapacity10 scsi command.

Changes committed:

	modified:   cmd/probe/capacityprobe.go
	modified:   cmd/probe/smartprobe.go
	modified:   pkg/smart/common.go
	modified:   pkg/smart/scsicommands.go

### DIFF
--- a/pkg/smart/common.go
+++ b/pkg/smart/common.go
@@ -59,23 +59,79 @@ func detectSCSIType(name string) (Dev, error) {
 	return &device, nil
 }
 
-// readDeviceCapacity sends a SCSI READ CAPACITY(16) command to a device and
-// returns the capacity in bytes.
-func (d *SCSIDev) readDeviceCapacity() (uint64, error) {
-	respBuf := make([]byte, 8)
+// sendReadCap10 is used to send a SCSIReadCapacity10 scsi command to
+// get logical block address and logical size of a disk
+func (d *SCSIDev) sendReadCap10() (uint32, uint32, error) {
+	respBuf10 := make([]byte, 8) // respBuf for read capacity(10) scsi cmd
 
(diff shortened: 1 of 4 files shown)


pkg/smart/common.go
```diff
@@ -59,23 +59,79 @@ func detectSCSIType(name string) (Dev, error) {
...
 
-// readDeviceCapacity sends a SCSI READ CAPACITY(16) command to a device and
-// returns the capacity in bytes.
+// sendReadCap10 is used to send a SCSIReadCapacity10 scsi command to
+// get logical block address and logical size of a disk
+func (d *SCSIDev) sendReadCap10() (uint32, uint32, error) {
+	respBuf10 := make([]byte, 8) // respBuf for read capacity(10) scsi cmd
+
+	// Populate the CDB10 required fields to send SCSIReadCapacity10 command
+	cdb10 := CDB16{SCSIReadCapacity10}
```
kmova: Should this be CDB10?

kmova:
Closing this PR in favor of #150

pkg/smart/common.go
sagarkrsd: Right @kmova, it should be CDB10.
It will work with CDB16 also but better to keep CDB10 here.

Issue #147, opened 2018-10-16
[BUG] Disk Capacity Mismatch: NDM can not recognize the disk spaces of more than 2TB 


zacharyzhao:
Hi, OpenEBS team,

I'm trying OpenEBS v0.7 and hitting an NDM Disk Capacity Mismatch Issue, it seems that NDM cannot recognize the disk spaces of more than 2TB.

For Example, my `/dev/sdb` has 36TB in total, but I got it's capacity information of `2TB` through `kubectl get disk`.

![wx20181016-140148](https://user-images.githubusercontent.com/13870117/46995677-4c24cd00-d14c-11e8-8d51-5a54588b4a64.png)

And under @kmova 's help, I know this issue is caused by incorrect `CDB` type in [scsicommands.go](https://github.com/openebs/node-disk-manager/blob/master/pkg/smart/scsicommands.go)

Please check it, thanks very much!

PR #150, opened 2018-10-17
Fix readDeviceCapacity method (#147)

patheticpat:
`readDeviceCapacity` used to send only a SCSI READ CAPACITY(10) command.
If this reports an overflow, issue an additional READ CAPACITY(16)
command to get the correct device capacity.

Signed-off-by: Michael Kaiser <patheticpat@googlemail.com>

### DIFF
--- a/pkg/smart/common.go
+++ b/pkg/smart/common.go
@@ -17,10 +17,13 @@ package smart
 
 import (
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"regexp"
 	"strings"
 )
+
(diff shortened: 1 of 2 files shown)


richardelling:
I prefer an approach where the opcodes are read (see sg_opcodes) and then the largest read_capacities is used.

kmova:
@patheticpat - Thank for this PR. Just wondering if you would like to address the above comment within this same PR.

patheticpat:
@kmova Sure, I'll look into it tomorrow.

Issue #164, opened 2018-10-23
Refactor code to send the SCSI CDB commands with highest supported version


kmova:
The current code for determining capacity issues a CDB10 command and if there is a over-flow issues a CDB16 command. Instead, it would be good to have a mechanism to determine the highest version supported upfront and use that for subsequent commands.

patheticpat:
I just pushed a PR (#174) addressing this issue.

akhilerm:
Closing as the changes are no more required as per discussions in #174

PR #174, opened 2018-10-30
Refactor readDeviceCapacity method (#164)

patheticpat:
Instead of sending a `READ CAPACITY(10)` command, checking for an
overflow, and possibly retrying, determine if the `READ CAPACITY(16)`
command is supported upfront and use it accordingly.

Signed-off-by: Michael Kaiser <patheticpat@googlemail.com>

### DIFF
--- a/pkg/smart/common.go
+++ b/pkg/smart/common.go
@@ -17,13 +17,10 @@ package smart
 
 import (
 	"encoding/binary"
-	"errors"
 	"fmt"
 	"regexp"
 	"strings"
 )
-
-var errReadCapacityOverflow = errors.New("READ CAPACITY (10) overflow")
 
 // detectBusType detects the type of bus such as SCSI,Nvme,etc based on the device path given
 func detectBusType(dname string) string {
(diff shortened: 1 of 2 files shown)


pkg/smart/common.go
```diff
@@ -62,14 +59,39 @@ func detectSCSIType(name string) (Dev, error) {
...
+	cdb[2] = 0x02
+	cdb[3] = SCSIReadCapacity16
+	binary.BigEndian.PutUint16(cdb[4:], uint16(SCSIReadCapacitySA))
+	binary.BigEndian.PutUint32(cdb[6:], uint32(len(respBuf)))
+
+	if err := d.sendSCSICDB(cdb[:], &respBuf); err != nil {
+		return false, err
+	}
+
+	switch respBuf[1] & 0x07 {
```
richardelling: it would be helpful for reviewers to know what these values mean, perhaps a comment or as a more descriptive constant

pkg/smart/common.go
```diff
@@ -62,14 +59,39 @@ func detectSCSIType(name string) (Dev, error) {
...
-// overflow, it retries a SCSI READ CAPACITY(16) command.
+// isReadDeviceCapacity16Supported checks if the SCSI READ CAPACITY(16)
+// command is supported by the device.
+func (d *SCSIDev) isReadDeviceCapacity16Supported() (bool, error) {
+	respBuf := make([]byte, 20)
+
+	cdb := CDB12{}
+	cdb[0] = SCSIReportSupportedOpCodes
+	cdb[1] = SCSIReportSupportedOpCodesSA
+	cdb[2] = 0x02
```
sagarkrsd: Can we use 0x02 as a constant and may be have some description (or the constant could itself be descriptive) why we need to populate this particular field ?

pkg/smart/common.go
```diff
@@ -62,14 +59,39 @@ func detectSCSIType(name string) (Dev, error) {
...
+		return true, nil
+	default:
+		return false, nil
+	}
+}
+
+// readDeviceCapacity returns the devices capacity in bytes.
 func (d *SCSIDev) readDeviceCapacity() (uint64, error) {
-	if DeviceCapacity, err := d.readDeviceCapacity10(); err == errReadCapacityOverflow {
+	if supported16, err := d.isReadDeviceCapacity16Supported(); err != nil {
```
vishnuitta: on `err`, would it be good to do readDeviceCapacity10()?

pkg/smart/scsicommands.go
```diff
@@ -23,23 +23,28 @@ import (
...
-// CDB6 is an array of 6 byte
+// CDB6 is an array of 6 bytes
 type CDB6 [6]byte
 
-// CDB10 is an array of 10 byte
+// CDB10 is an array of 10 bytes
 type CDB10 [10]byte
 
-// CDB16 is an array of 16 byte
+// CDB12 is an array of 10 bytes
```
vishnuitta: CDB12 is an array of 12 bytes

patheticpat: Argh, of course, stupid copy and paste mistake

pkg/smart/common.go
patheticpat: I don't know in which scenario `isReadDeviceCapacity16Supported()` might fail, but `readDeviceCapacity10()` still work. But if it does, I assume `readDeviceCapacity10()` would need to check for the capacity overflow again.

vishnuitta:
3.25 of https://www.seagate.com/staticfiles/support/disc/manuals/scsi/100293068a.pdf says that - it is fine to do ReadCapacity16 if `RETURNED LOGICAL BLOCK ADDRESS` is `0xFFFFFFFF`.

patheticpat:
@vishnuitta Yes, and that's exactly what the original code (#150) did. I was about to comment that in my opinion, it is not worth adding all this complexity.

vishnuitta:
given that current code is following the protocol, I'm of opinion that these changes may not be required. Would like to hear from others.

richardelling:
The most correct method is to get the opcodes and use the appropriate opcode.
Next best approach is try read_capacity(16) and if that fails, read_capacity(10).
Least efficient is read_capacity(10) then read_capacity(16).

pkg/smart/common.go
sagarkrsd: No issue @patheticpat. It's ok if we even give some description for the field instead of having the constant.

kmova:
Base on the above comments from @patheticpat @vishnuitta @richardelling - closing this PR in favor of the current behavior.

Issue #182, opened 2018-12-04
fix broken Makefile rule "vet" to do "go vet"


michael-db:
With the Makefile as it stands
make vet
does nothing.
It is intended to run "go vet -composites" but is broken because of the two lines following it.

PR #184, opened 2018-12-04
fix broken Makefile rule "vet" to do "go vet" #182

michael-db:
Remove extraneous lines following the "vet:" line.

Signed-off-by: Michael Breen <8309342+michael-db@users.noreply.github.com>


kmova:
Thanks @michael-db

PR #185, opened 2018-12-06
[TA4346] feat: Adding some more stats using OpenSeaChest library.

satbirchhikara:
This PR is to add more information in Disk-CR using newly integrated OpenSeaChest library.
Red colored fields are newly added:-
```
satbir@node-disk-manager:$ kubectl describe disk disk-2cd7034c2b3329392ae07eb83bd15fad
Name:         disk-2cd7034c2b3329392ae07eb83bd15fad
Namespace:    
Labels:       kubernetes.io/hostname=gke-satbir-ndm-default-pool-32f04232-28rl
              ndm.io/disk-type=disk
Annotations:  <none>
API Version:  openebs.io/v1alpha1
Kind:         Disk
Metadata:
  Cluster Name:        
  Creation Timestamp:  2018-11-27T09:46:34Z
  Generation:          1
  Resource Version:    841260
  Self Link:           /apis/openebs.io/v1alpha1/disks/disk-2cd7034c2b3329392ae07eb83bd15fad
  UID:                 53b9c334-f229-11e8-b598-42010a8000c4
Spec:
  Capacity:
    Physical Sector Size:  4096   <<<Newly added
    Logical Sector Size:   512
    Storage:               10737418240
  Details:
    Disk Type:          disk
    Rotation Rate:      0             <<<Newly added
    Compliance:         SPC-4
    Firmware Revision:  1   
    Model:              PersistentDisk
    Serial:             satbir-ndm-disk-2
    Vendor:             Google
  Devlinks:
    Kind:  by-id
    Links:
      /dev/disk/by-id/scsi-0Google_PersistentDisk_satbir-ndm-disk-2
      /dev/disk/by-id/google-satbir-ndm-disk-2
    Kind:  by-path
    Links:
      /dev/disk/by-path/virtio-pci-0000:00:03.0-scsi-0:0:3:0
  Path:  /dev/sdc
Stats:  <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<Newly added
  Device Utilization Rate:  0
  Disk Temperature:
    Current Temperature:   0
    Highest Temperature:   0
    Lowest Temperature:    0
  Percent Endurance Used:  0
  Total Bytes Read:        0
  Total Bytes Written:     0
Status:
  State:  Active
Events:   <none> 
```


pkg/seachest/seachest.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+func (I *Identifier) GetRotationRate(driveInfo *C.driveInformationSAS_SATA) uint16 {
+	if driveInfo.rotationRate > 1 {
+		return ((uint16)(driveInfo.rotationRate))
+	}
+	return 0
+}
+
+func (I *Identifier) IsRotational(driveInfo *C.driveInformationSAS_SATA) string {
```
richardelling: you're getting closer, I think this should return a boolean. Given the way we often see this in the field, I think we can safely say:
```
if driveInfo.rotationRate > 1 {
  return True
}
return false
```

satbirchhikara: Since we are filling drive type "SSD" or "HDD", I renamed function to driveType. This is something inline with OpenSeaChest Utility. They classify deive exactly the way we are doing here.

cmd/probe/smartprobe.go
```diff
@@ -34,7 +34,7 @@ type smartProbe struct {
 
 const (
 	smartConfigKey     = "smart-probe"
-	smartProbePriority = 2
+	smartProbePriority = 3
```
vishnuitta: should we have similar change for udevProbePriority as well from 1 to 2?

vishnuitta: ok.. seaChestProbePriority is set to 2.. so, udevProbePriority should be fine then

.travis.yml
```diff
@@ -1,3 +1,4 @@
+dist: xenial
 language: go
```
pawanpraka1: we should probably compile it in a docker. We need dapper for ndm probably in future PR.

cmd/probe/probe.go
```diff
@@ -28,6 +28,7 @@ const (
 
 // RegisteredProbes contains register function of probes which we want to register
 var RegisteredProbes = []func(){
```
pawanpraka1: for better code readability please reorder this list in the priority order (next PR).

Issue #194, opened 2019-01-07
NDM pods experienced panic while I attach a disk. 


imazik:
**What happened**:
node-disk-manager was running in my cluster and I attached on disk in my node. Then it experienced panic.

**How to reproduce it (as minimally and precisely as possible)**:
Run ndm in your cluster then attach a disk in any node.

**Anything else we need to know?**:
I am running 1.12 k8s cluster.
```
shovan@instance:~$ kubectl version
Client Version: version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.4", GitCommit:"f49fa022dbe63faafd0da106ef7e05a29721d3f1", GitTreeState:"clean", BuildDate:"2018-12-14T07:10:00Z", GoVersion:"go1.10.4", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.4", GitCommit:"f49fa022dbe63faafd0da106ef7e05a29721d3f1", GitTreeState:"clean", BuildDate:"2018-12-14T06:59:37Z", GoVersion:"go1.10.4", Compiler:"gc", Platform:"linux/amd64"}
```
dev path `/dev/sdX` is not present while seachest probe trying to get details of that disk.

logs of ndm pod - 
```
I0107 04:19:39.275215       8 event.go:44] processing new event for disk-125ef30ad47c7583e17bab21612b7f29 action type add
I0107 04:19:39.283678       8 eventhandler.go:49] processing details for disk-125ef30ad47c7583e17bab21612b7f29
I0107 04:19:39.283938       8 probe.go:107] details filled by udev probe
open: No such file or directory
E0107 04:19:39.284105       8 seachest.go:84] Unable to get device info for device:%s with error:%s/dev/sdbFailed
E0107 04:19:39.284115       8 seachestprobe.go:100] 1
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x100 pc=0xef286b]

goroutine 9 [running]:
panic(0x109f140, 0x1ea33d0)
	/usr/local/go/src/runtime/panic.go:556 +0x2cb fp=0xc0005a7a28 sp=0xc0005a7998 pc=0x43ab3b
runtime.panicmem()
	/usr/local/go/src/runtime/panic.go:82 +0x5e fp=0xc0005a7a48 sp=0xc0005a7a28 pc=0x439b7e
runtime.sigpanic()
[...]

PR #198, opened 2019-01-31
Error handling in seachest probe

akhilerm:
If NDM is run in unprivileged mode, Seachest probe was causing the pod into `CrashLoopBackOff` state. This was because Seachest was having insufficient permissions to fetch the details. The PR aborts filling the details by Seachest probe if its not able to get the `driveInfo` pointer.


akhilerm:
This PR fixes issue #194

PR #201, opened 2019-02-21
[WIP][US4295] feat(diskCR): detecting change/delete udev events

akhilerm:
Add/delete events for partitions will be now monitored and the corresponding changes will be updated in the diskCR. When a partition is deleted, the details of the last partition will be removed from the diskCR instead of removing the exact partition. This is because a unique id to detect partitions is yet to be implemented. Changes in disk/devices (add/delete) will be now monitored through udev events and corresponding updates will be made in the diskCR.

### DIFF
--- a/cmd/ndm_daemonset/probe/eventhandler.go
+++ b/cmd/ndm_daemonset/probe/eventhandler.go
@@ -37,8 +37,8 @@ type ProbeEvent struct {
 	Controller *controller.Controller
 }
 
-// addDiskEvent fill disk details from different probes and push it to etcd
-func (pe *ProbeEvent) addDiskEvent(msg controller.EventMessage) {
+// addEvent fill disk/device details from different probes and push it to etcd
+func (pe *ProbeEvent) addEvent(msg controller.EventMessage) {
 	diskList, err := pe.Controller.ListDiskResource()
 	if err != nil {
 		glog.Error(err)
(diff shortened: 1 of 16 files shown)


cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -98,7 +94,17 @@ func (pe *ProbeEvent) deleteDiskEvent(msg controller.EventMessage) {
...
 			continue
 		}
-		pe.Controller.DeactivateDisk(*oldDr)
+		if diskDetails.DiskType == libudevwrapper.UDEV_DISK {
+			pe.Controller.DeactivateDisk(*oldDr)
+		} else if diskDetails.DiskType == libudevwrapper.UDEV_PARTITION {
+			glog.Info("Removing Partition Data from ", diskDetails.ProbeIdentifiers.Uuid)
+			newDrCopy := oldDr.DeepCopy()
+			if len(newDrCopy.Spec.PartitionDetails) > 0 {
+				// Currently the details of last partition will be removed
```
satbirchhikara: How you would like to handle deleting a specific partition, please explain it. Also put TODO in comment section so that it is in our RADAR to fix.

akhilerm: For deleting a specific partition we need to have a UID for each partition, so that it can be selected and deleted.

satbirchhikara: We can add partition UID into partition info so that we would be able to identify right partition and delete that partition only. 
Regarding how to generate UID of partition, I am think if we can appen partition number to disk UID. For example :-
disk(sdb) UID = xxxxxxAAAAA
Partition (sdb1) UID = xxxxxxAAAAA-1
Partition (sdb2) UID = xxxxxxAAAAA-2
That way when a delete event comes, we can findout which partition info need to be removed from disk-partition info.

PR #207, opened 2019-02-26
[US4256] feat(operator-sdk): Moving "cmd/*"  to "cmd/ndm_daemonset/" for operator-sdk

satbirchhikara:
This change is required to accomodate two main.go for operator-sdk. So exisitng files under dir "cmd" has been moved to new dir "cmd/ndm_daemonset".


hack/build.sh
```diff
@@ -66,7 +66,7 @@ build(){
                 -X main.CtlName='${CTLNAME}' \
                 -X github.com/openebs/node-disk-manager/pkg/version.Version=${VERSION}" \
                 -o $output_name \
-               ./cmd
+               ./cmd/ndm_daemonset
```
kmova: @satbirchhikara -- can this be taken as an argument -- just like the CTLNAME so that we don't need to create one more build script for operator.

satbirchhikara: Done. Please have a look and let me know.

PR #217, opened 2019-03-01
[US4256] feat (operator): Node Disk Operator using operator-sdk.

satbirchhikara:
Node Disk Operator feature implemented using operator-sdk which include Disk, Device and Devicerequest custom resource definition and controller for all custom resources.

This PR has auto generated code for device and devicerequest. Disk CR has fields defined/declared in it because existing Unit test case should pass and for that I have added fields in Disk CR.


cmd/ndm_daemonset/controller/diskstore.go
```diff
@@ -17,18 +17,21 @@ limitations under the License.
...
 	"github.com/openebs/node-disk-manager/pkg/util"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // CreateDisk creates the Disk resource in etcd
 func (c *Controller) CreateDisk(dr apis.Disk) {
 	drCopy := dr.DeepCopy()
-	cdr, err := c.Clientset.OpenebsV1alpha1().Disks().Create(drCopy)
+	err := c.Clientset.Create(context.TODO(), drCopy)
```
kmova: What is a TODO? Can we add a comment about this of what is it supposed to do?

kmova:
@satbirchhikara -- I see a change in the directory naming from `openebs.io` to `openebs` for accessing the API. Is this an operator-sdk enforced change?

cmd/ndm_daemonset/controller/diskstore.go
satbirchhikara: Run type Client APIs need some sort of context. using context.TODO() is way to pass empty context. More on it can be found here ....https://www.notion.so/context-TODO-vs-context-Background-d5224e27ff724a33a79cb4e03a5eb333

We need to take this Context exercise in all modules of OpenEBS sometime soon, it will be fixed in that context.

satbirchhikara:
> Since we have removed CITF dependencies, the setting up of minikube in `.travis.yml` is no more required. Those [lines](https://github.com/openebs/node-disk-manager/blob/master/.travis.yml#L31) which perform the setup of minikube binary initially should be removed.

Thanks Akhil for pointing it out. I would like to take this comment for next (Pending) PR. Infact I have done these changes in my next PR.

satbirchhikara:
> @satbirchhikara -- I see a change in the directory naming from `openebs.io` to `openebs` for accessing the API. Is this an operator-sdk enforced change?

Yes, I gave "openebs.io" but it created "openebs"....its being forced on us by operator-sdk.

cmd/ndm_daemonset/controller/diskstore.go
satbirchhikara: context.TODO() is the way to pass empty context when application/module is not using any context. This is the operator-sdk way of doing things.

PR #223, opened 2019-03-25
 [US4449] feat(NDM-Pod Liveness check): ndm-liveness implementation.

satbirchhikara:
[US4449] feat(NDM-Pod Liveness check): ndm-liveness implementation.
    
    node-disk-manager DaemonSet liveness check implemented using
    ndmconfig CR & Controller. For each ndm DaemonSet pod, a ndmconfig
    would be created, pushed to etcd. ndmconfig CR would be having owenref
    set to pod. ndmconfig controller keep a watch on ndmconfig CR, when pod
    crash, it will try to deleted ndmconfig and as a part of deletion,
    ndmconfig controller will mark device/disks Unknown for that node and
    deleted ndmconig.
    

kmova:
Keeping this functionality on hold - till we get the initial implementation of BD/BDC used in the openebs cstor and local pv provisioners.

PR #228, opened 2019-03-29
fix(yaml): added selinux options for ndm daemon-set

akhilerm:
While deploying NDM to systems in which selinux was enabled, it was producing OCI runtime error. This was because selinux was not allowing to run containers in privileged mode. By setting the required SELinux labels to the daemonset container, the pod can be run in privileged mode without turning off selinux.


kmova:
Closing this in favor of https://github.com/openebs/node-disk-manager/pull/238

PR #230, opened 2019-04-08
fix(udev): added nil check while getting udev devices

akhilerm:
If `libudev` returns an invalid address/nil pointer while trying to create a udev struct from syspath, the error was not checked or logged. 

This PR enables logging and  error checking,  and aborts fetching details using udev probe if libudev returns nil.

### DIFF
--- a/cmd/ndm_daemonset/probe/udevprobe.go
+++ b/cmd/ndm_daemonset/probe/udevprobe.go
@@ -171,7 +171,7 @@ func (up *udevProbe) scan() error {
 func (up *udevProbe) FillDiskDetails(d *controller.DiskInfo) {
 	udevDevice, err := newUdevProbeForFillDiskDetails(d.ProbeIdentifiers.UdevIdentifier)
 	if err != nil {
-		glog.Error(err)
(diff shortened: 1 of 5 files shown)


akhilerm:
Fixes [#2498](https://github.com/openebs/openebs/issues/2498)

cmd/ndm_daemonset/probe/udevprobe.go
```diff
@@ -169,7 +169,11 @@ func (up *udevProbe) scan() error {
 
 // fillDiskDetails filles details in diskInfo struct using probe information
 func (up *udevProbe) FillDiskDetails(d *controller.DiskInfo) {
-	udevDevice := newUdevProbeForFillDiskDetails(d.ProbeIdentifiers.UdevIdentifier)
+	udevDevice, err := newUdevProbeForFillDiskDetails(d.ProbeIdentifiers.UdevIdentifier)
+	if err != nil {
+		glog.Error(err)
```
imazik: can we add more log here like value of identifier

akhilerm: Done.

PR #234, opened 2019-04-10
fix(controller): fixed client init and CR update

akhilerm:
runtime client is now created using client.New() instead of from the manager. update on the disk/device CR now takes the latest resource version and updates the resource

### DIFF
--- a/cmd/ndm_daemonset/controller/controller.go
+++ b/cmd/ndm_daemonset/controller/controller.go
@@ -135,7 +135,7 @@ func NewController(kubeconfig string) (*Controller, error) {
 		return controller, err
 	}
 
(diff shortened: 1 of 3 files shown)


akhilerm:
In operator-sdk, `controller-runtime`'s client is used for cluster interaction. By default SDK relies on `manager.Manager` to create a `client.Client`. But this client reads from the cache and writes to the API server, thus reducing read load on the API server. But this has issues that information regarding resource may not be updated in the cache. Due to this issue, the disk CR was not updated in the cache and when we try to `Get()` the resource it returned an error. In such cases, we can write a custom client using `client.New()`, which serves all requests from the API server.

PR #236, opened 2019-04-15
feat (BDD) : NDM integration testing

akhilerm:
1) Added YAMLs for integration testing
2) Udev functions for simulating devices. The udev functions could be used to trigger fake udev events so that ndm detects the events and updates the CRs. functions will be required to create a fake disk,
trigger add/remove/change events
3) sanity tests will be done in a single node minikube cluster and functions for interfacing with the cluster is included.The cluster start and stop is not included as this will be done by the VM. Status of cluster, and generation of kube-config files will be checked
4) package to query kube-API server and retrieve results in a predefined format is included. The various resources like configmap, CRD etc can also be created using this package. These resources will be created using the YAMLs
5) The basic tests include:
        a) running NDM and checking for pod
        b) checking for sparse disks
        c) check for external disk already attached
        d) check for external disk attached at runtime


integration_tests/k8s/k8sclient.go
```diff
@@ -0,0 +1,72 @@
...
+
+type k8sClient struct {
+	ClientSet     *kubernetes.Clientset
+	APIextClient  *apiextensionsclient.Clientset
+	RunTimeClient client.Client
+}
+
+// GetClientSet generates the client-set from the config file on the host
+// 3 clients are generated by the function:
+//		1. Client-set from client-go which is used for operations on pods/nodes and
```
kmova: n_: needs indent fix

integration_tests/sanity/k8s_tests.go
```diff
@@ -0,0 +1,75 @@
...
+	DiskName = "disk-"
+	// ActiveState stores the active state of Disk/Device resource
+	ActiveState = "Active"
+	// InactiveState stores the deactivated state of Disk/Device resource
+	InactiveState = "Inactive"
+	// DiskImageSize is the default file size used while creating backing image
+	DiskImageSize = 1073741824
+)
+
+var _ = Describe("NDM Basic Tests", func() {
```
kmova: We can this block as "NDM Setup Tests"

integration_tests/sanity/customresources_test.go
```diff
@@ -0,0 +1,140 @@
...
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/openebs/node-disk-manager/integration_tests/k8s"
+	"github.com/openebs/node-disk-manager/integration_tests/udev"
+	"strings"
+)
+
+var _ = Describe("Custom Resource Tests", func() {
```
kmova: This block can be called "Device Discovery Tests"

integration_tests/sanity/ndm_integration_test.go
```diff
@@ -0,0 +1,41 @@
...
+
+// Initialize the minikube cluster
+var _ = BeforeSuite(func() {
+	var err error
+	Expect(minikube.IsUpAndRunning()).To(BeTrue())
+	err = minikube.WaitForMinikubeToBeReady()
+	Expect(err).ShouldNot(HaveOccurred())
+})
+
+var _ = Describe("Minikube Cluster", func() {
```
kmova: This block can be called "Verify Minikube Setup"

integration_tests/k8s/k8sapi.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	for _, pod := range podList.Items {
+		pods[pod.Name] = string(pod.Status.Phase)
+	}
+	return pods, nil
+}
+
+// GetNodes returns list of all nodes(node name) in the cluster along with
+// their status
+func GetNodes(clientSet *kubernetes.Clientset) (map[string]string, error) {
```
imazik: GetNodes() should return node.
May be you need to change function name and signature here

integration_tests/k8s/k8sapi.go
```diff
@@ -0,0 +1,140 @@
...
+		return nil, err
+	}
+	for _, node := range nodeList.Items {
+		nodes[node.Name] = string(node.Status.Phase)
+	}
+	return nodes, nil
+}
+
+// GetDiskList returns list of DiskCR in the cluster
+func GetDiskList(clientSet k8sClient) (*v1alpha1.DiskList, error) {
```
imazik: `ListDisk()` may be good here.

integration_tests/k8s/k8sapi.go
```diff
@@ -0,0 +1,140 @@
+package k8s
```
imazik: we should bind all the function with a struct thatt contains `*kubernetes.Clientset`.
Better to avoid functions please use methods

integration_tests/k8s/k8sgen.go
```diff
@@ -0,0 +1,111 @@
...
+	ClusterRoleYAML        NDMYaml = "../yamls/clusterrole.yaml"
+	ClusterRoleBindingYAML NDMYaml = "../yamls/clusterrolebinding.yaml"
+	DiskCRDYAML            NDMYaml = "../yamls/diskCR.yaml"
+	DeviceCRDYAML          NDMYaml = "../yamls/deviceCR.yaml"
+	DeviceRequestCRDYAML   NDMYaml = "../yamls/deviceRequestCR.yaml"
+	DaemonSetYAML          NDMYaml = "../yamls/daemonset.yaml"
+)
+
+// GetConfigMap generates the ConfigMap object for NDM from the yaml file
+func GetConfigMap() (v1.ConfigMap, error) {
```
imazik: can we change it to 
```go
func GetConfigMapFromYaml(yaml string) (v1.ConfigMap, error)
```

integration_tests/k8s/k8sgen.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	err = yaml.Unmarshal([]byte(yamlstring), &configMap)
+	if err != nil {
+		return configMap, err
+	}
+	return configMap, nil
+}
+
+// GetServiceAccount generates the ServiceAccount object from the yaml file
+func GetServiceAccount() (v1.ServiceAccount, error) {
```
imazik: same comment.

integration_tests/k8s/k8sgen.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	err = yaml.Unmarshal([]byte(yamlstring), &serviceAccount)
+	if err != nil {
+		return serviceAccount, err
+	}
+	return serviceAccount, nil
+}
+
+// GetClusterRole generates the ClusterRole object from the yaml file
+func GetClusterRole() (rbacv1beta1.ClusterRole, error) {
```
imazik: same comment.

integration_tests/k8s/k8sgen.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	err = yaml.Unmarshal([]byte(yamlstring), &clusterRole)
+	if err != nil {
+		return clusterRole, err
+	}
+	return clusterRole, nil
+}
+
+// GetClusterRoleBinding generates the ClusterRoleBinding object from the yaml file
+func GetClusterRoleBinding() (rbacv1beta1.ClusterRoleBinding, error) {
```
imazik: same comment.

integration_tests/k8s/k8sgen.go
```diff
@@ -0,0 +1,111 @@
...
+	err = yaml.Unmarshal([]byte(yamlstring), &clusterRoleBinding)
+	if err != nil {
+		return clusterRoleBinding, err
+	}
+	return clusterRoleBinding, err
+}
+
+// GetCustomResourceDefinition generates the CustomResourceDefinition object from the specified
+// YAML file
+func GetCustomResourceDefinition(crdyaml NDMYaml) (apiextensionsv1beta1.CustomResourceDefinition, error) {
```
imazik: same comment.

integration_tests/k8s/k8sgen.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	err = yaml.Unmarshal([]byte(yamlString), &customResourceDefinition)
+	if err != nil {
+		return customResourceDefinition, err
+	}
+	return customResourceDefinition, err
+}
+
+// GetDaemonSet generates the NDM DaemonSet object from the yaml file
+func GetDaemonSet() (v1beta1.DaemonSet, error) {
```
imazik: same comment.

integration_tests/k8s/k8sapi.go
imazik: If we are creating a wrapper over k8s client exposed functions need to think about packaging.

imazik: These are the core functionality instead of putting it here we can add it inside `pkg` so that other can use it.
If you put it here it is accessible in other packages but import path will not be meaningful.

integration_tests/k8s/k8sclient.go
```diff
@@ -0,0 +1,72 @@
+package k8s
```
imazik: can we have a client common client package so that we can use it in both ndm code and it.

akhilerm: @imazik , currently we will keep the BDD code and NDM business logic code separate. If they converge on the same methods, then we will move into `pkg`

integration_tests/k8s/k8sapi.go
akhilerm: @imazik , currently we will keep the BDD code and NDM business logic code separate. If they converge on the same methods, then we will move into pkg

integration_tests/k8s/k8sapi.go
akhilerm: Done.

integration_tests/k8s/k8sapi.go
akhilerm: Function signature changed.

Issue #237, opened 2019-04-17
Run NDM on hosts with SELinux enabled


akhilerm:
NDM should be able to run in systems with selinux enabled. Currently it is failing because `/proc/1/mounts` of host cannot be mounted inside container. This results in the os-disk-filter not being able to work correctly.

**What is the expected behavior?**
OS-disk filter should be able to detect the mounted disks correctly.

akhilerm:
NDM can be run on hosts with SELinux enabled, from release `v0.3.5`

PR #238, opened 2019-04-17
fix(code,yaml): fix to mount the /proc directory of host

akhilerm:
`/proc/1/mounts` cannot be mounted into container in systems with selinux enabled. But `/proc` directory can be mounted. This fix mounts the `/proc` directory to `/host/proc` inside container and thus the mounts file will be available at `/host/proc/1/mounts`.


akhilerm:
Fixes #237

PR #256, opened 2019-05-28
fix(blockdevice): deactivate block device when Device is removed

### LINKED ISSUE #255: BlockDevice is not deactivated when the device is removed
**What happened**:
NDM was installed using `ndm-operator.yaml`. A single physical disk is attached to the node. Block Device and Disk resource are created for the newly attached disk. When removing the physical disk, the diskCR is getting deactivated. But the BlockDevice CR is not getting deactivated.

**What you expected to happen**:
The status of BlockDevice should be updated to `Inactive`

**How to reproduce it (as minimally and precisely as possible)**:
1. Install NDM from `ndm-operator.yaml`
2. Attach a physical disk
3. Check the status of BlockDevice CR
4. Remove the physical disk
5. Check the status of BlockDevice CR

**Anything else we need to know?**:
CI image was used for installing NDM

akhilerm:
BlockDevice will get deactivated when the corresponding device is removed from the host.
Fixes #255 

### DIFF
--- a/cmd/ndm_daemonset/probe/eventhandler.go
+++ b/cmd/ndm_daemonset/probe/eventhandler.go
@@ -102,11 +102,16 @@ func (pe *ProbeEvent) addDiskEvent(msg controller.EventMessage) {
 	}
 }
 
-// deleteDiskEvent deactivate disk resource using uuid from etcd
+// deleteDiskEvent deactivate disk/blockdevice resource using uuid from etcd
 func (pe *ProbeEvent) deleteDiskEvent(msg controller.EventMessage) {
-	diskList, err := pe.Controller.ListDiskResource()
-	if err != nil {
-		glog.Error(err)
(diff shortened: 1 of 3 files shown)


cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -102,11 +102,16 @@ func (pe *ProbeEvent) addDiskEvent(msg controller.EventMessage) {
 	}
 }
 
-// deleteDiskEvent deactivate disk resource using uuid from etcd
+// deleteDiskEvent deactivate disk/blockdevice resource using uuid from etcd
```
kmova: @akhilerm -- how about splitting this function into two - for block device and disk?

akhilerm: @kmova . The function is called when we receive a udev Remove Event. At that point, we do not have a Disk/BlockDevice separation.

PR #262, opened 2019-06-05
fix(eventhandler): fix blockdevice not getting deactivated

akhilerm:
The UUID used earlier was of the format 'disk-xxx', this need to be changed to `blockdevice-xxx`


cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -123,7 +124,9 @@ func (pe *ProbeEvent) deleteBlockDevice(msg controller.EventMessage) bool {
 	}
 	ok := true
 	for _, diskDetails := range msg.Devices {
-		oldBDResource := pe.Controller.GetExistingBlockDeviceResource(bdList, diskDetails.ProbeIdentifiers.Uuid)
+		UUID := strings.TrimPrefix(diskDetails.ProbeIdentifiers.Uuid, libudevwrapper.NDMDiskPrefix)
```
imazik: can we have some function that will convert 
`disk` to `block-device`  so that we can reuse it.

PR #273, opened 2019-06-20
fix(cleaner): change dd to wipefs

akhilerm:
Fixed high cpu/io utilization of cleanup job. Using `dd` caused heavy load on the CPU, and prevented other pods from getting scheduled on the node. It also caused kube-api server crash at times. By using `wipefs` instead of `dd` only the file-system signatures will be erased. This process is much faster than writing zeros, since we are only working with certain block(superblocks).

### DIFF
--- a/ndm-operator.yaml
+++ b/ndm-operator.yaml
@@ -270,4 +270,4 @@ spec:
             - name: OPERATOR_NAME
               value: "node-disk-operator"
             - name: CLEANUP_JOB_IMAGE
-              value: "quay.io/openebs/openebs-tools:3.8"
+              value: "openebs/linux-utils:3.9"
--- a/ndm-operator.yaml
+++ b/ndm-operator.yaml
@@ -270,4 +270,4 @@ spec:
             - name: OPERATOR_NAME
               value: "node-disk-operator"
             - name: CLEANUP_JOB_IMAGE
-              value: "openebs/linux-utils:3.9"
+              value: "quay.io/openebs/linux-utils:3.9"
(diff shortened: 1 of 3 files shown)


pkg/cleaner/jobcontroller.go
```diff
@@ -74,13 +73,13 @@ func NewCleanupJob(bd *v1alpha1.BlockDevice, volMode VolumeMode, namespace strin
...
-		blockCount := bd.Spec.Capacity.Storage / 1024 / 1024
-		count := "count=" + strconv.FormatUint(blockCount, 10)
-		jobContainer.Command = getCommand(BlockCleanerCommand, input, output, blockSize, count)
+		// wipefs erases the filesystem signature from the block
+		// -a    wipe all magic strings
+		// -f    force erasure
+		wipeCommand := BlockCleanerCommand + " -af " + bd.Spec.Path
+
+		jobContainer.Command = []string{"/bin/sh", "-c"}
+		jobContainer.Args = []string{wipeCommand}
```
prateekpandey14: `jobContainer.Command = getCommand(BlockCleanerCommand, "-af", bd.Spec.Path)`
to make it more cleaner ?

akhilerm: `getCommand` returns a slice of strings. We are passing a string to container args right.

ndm-operator.yaml
```diff
@@ -270,4 +270,4 @@ spec:
             - name: OPERATOR_NAME
               value: "node-disk-operator"
             - name: CLEANUP_JOB_IMAGE
-              value: "quay.io/openebs/openebs-tools:3.8"
+              value: "openebs/linux-utils:3.9"
```
kmova: change to quay.

PR #277, opened 2019-07-03
feat(operator): automate installation of CRDs from NDM operator.

akhilerm:
automate installation of NDM Custom Resources(Disk, BlockDevice, BlockDeviceClaim) from NDM operator.

### DIFF
--- a/integration_tests/yamls/clusterrole.yaml
+++ b/integration_tests/yamls/clusterrole.yaml
@@ -4,6 +4,19 @@ metadata:
   namespace: default
   name: openebs-ndm-operator
 rules:
-- apiGroups: ["*"]
-  resources: ["disks", "blockdevices", "blockdeviceclaims"]
-  verbs: ["*"]
+  - apiGroups: ["*"]
+    resources: ["disks", "blockdevices", "blockdeviceclaims", "pods", "services", "endpoints", "events", "configmaps", "secrets", "jobs"]
+    verbs:
+      - '*'
+  - apiGroups: ["apiextensions.k8s.io"]
+    resources: ["customresourcedefinitions"]
+    verbs:
+      - '*'
+  - apiGroups:
+      - openebs.io
+    resources:
+      - '*'
+      - blockdevices
+      - blockdeviceclaims
+    verbs:
+      - '*'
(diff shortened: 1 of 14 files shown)


pkg/apis/openebs/v1alpha1/crds.go
```diff
@@ -0,0 +1,90 @@
...
+	BlockDeviceClaimResourceListKind = "BlockDeviceClaimList"
+	// BlockDeviceClaimResourcePlural is the plural form used for block device claim
+	BlockDeviceClaimResourcePlural = "blockdeviceclaims"
+	// BlockDeviceClaimResourceShort is the short name used for block device claim CRD
+	BlockDeviceClaimResourceShort = "bdc"
+	// BlockDeviceClaimResourceName is the name of the block device claim resource
+	BlockDeviceClaimResourceName = BlockDeviceClaimResourcePlural + "." + GroupName
+)
+
+func buildCRD(name, kind, listKind, plural, short string, scope apiext.ResourceScope) *apiext.CustomResourceDefinition {
```
kmova: These CRDs are missing custom columns for display. Also in the future if we want to add a few more details like schema validation, this method via parameters can become a bottleneck. Either we can modify this to a builder patter or make use of the YAML string -> generic kubeclient to load the CRD into K8s, like we do in the maya/pkg/install.

kmova: Thanks for changing this!

integration_tests/yamls/clusterrole.yaml
```diff
@@ -4,6 +4,19 @@ metadata:
...
+    verbs:
+      - '*'
+  - apiGroups: ["apiextensions.k8s.io"]
+    resources: ["customresourcedefinitions"]
+    verbs:
+      - '*'
+  - apiGroups:
+      - openebs.io
+    resources:
+      - '*'
```
kmova: is `*` required when explicitly mentioning the below items.

integration_tests/yamls/clusterrole.yaml
```diff
@@ -4,6 +4,19 @@ metadata:
   namespace: default
   name: openebs-ndm-operator
 rules:
-- apiGroups: ["*"]
-  resources: ["disks", "blockdevices", "blockdeviceclaims"]
-  verbs: ["*"]
+  - apiGroups: ["*"]
+    resources: ["disks", "blockdevices", "blockdeviceclaims", "pods", "services", "endpoints", "events", "configmaps", "secrets", "jobs"]
```
kmova: Similar to below, can we split this into next line. use a uniform style.

PR #281, opened 2019-07-09
fix(operator, upgrade): handle CRD upgrade from operator

akhilerm:
This PR handles the upgrade scenarios of CRDs (disk, bd and bdc). When a new change is seen in the CRD object, operator will create a json patch and update the CRD object with the patch.

### DIFF
--- a/pkg/setup/create_crd.go
+++ b/pkg/setup/create_crd.go
@@ -47,10 +47,10 @@ func (sc Config) createCRD(crd *apiext.CustomResourceDefinition) error {
 			// This will also handle the upgrades of CRDs
 			patch, err := json.Marshal(crd)
 			if err != nil {
-				return fmt.Errorf("could not marshal new customResourceDefintion: %v", err)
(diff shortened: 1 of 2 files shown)


pkg/setup/create_crd.go
```diff
@@ -41,10 +43,15 @@ func (sc Config) createBlockDeviceClaimCRD() error {
...
 		if errors.IsAlreadyExists(err) {
-			// CRD is already present, no need to do anything
-			// In future can implement upgrade of CRDs here.
-			// For upgrade, a patch can be created which can then be
-			// used to upgrade the CRD
+			// if CRD already exists, we patch it with the new changes.
+			// This will also handle the upgrades of CRDs
+			patch, err := json.Marshal(crd)
+			if err != nil {
+				return fmt.Errorf("could not marshal new customResourceDefintion: %v", err)
```
kmova: Does the error message include the name of the CRD? If not it will be good to mention which crd is having the issue.

PR #283, opened 2019-07-11
fix(controller): add extra checks before removing finalizer

akhilerm:
To prevent accidental deletion of BDC, the BDC controller will now perform an additional check to make sure that the finalizer added by NDM will be the last one to be removed from the BDC object.

The new workflow will be:
- NDM Operator
1. NDM Operator will be watching all the BDCs
2. If DeletionTimeStamp is set on the BDC, then, Controller will check for the no.of finalizers on the BDC, if it has only one finalizer, then check whether NDM finalizer is present. If yes, the BD will be marked as released and the NDM finalizer will be removed.
3. NDM operator will make sure that, its finalizer is the last one to be removed from the BDC

- Maya (Owner of BDC)
1. Maya will create the BDC and add a finalizer to it.
2. When pool/localpv is deleted and cleanedup, this finalizer should be removed by maya, and will delete the BDC
3. Once deletion timestamp is set, NDM operator takes up the work.

This approach will give partial control of how the BDC should be deleted to the owner of BDC. If the user who creates the claim, want to prevent accidental deletion, he can add a finalizer but then it will be the responsibility of the user to remove it.

This approach also makes sure that BD will be marked as Released only when no one else is using the corresponding BDC.

### DIFF
--- a/pkg/controller/blockdeviceclaim/blockdeviceclaim_controller.go
+++ b/pkg/controller/blockdeviceclaim/blockdeviceclaim_controller.go
@@ -46,7 +46,7 @@ var log = logf.Log.WithName("controller_deviceclaim")
 
 const (
 	// BlockDeviceClaimFinalizer is the finalizer name for the block device claim
-	BlockDeviceClaimFinalizer = "blockdeviceclaim.finalizer"
+	BlockDeviceClaimFinalizer = "openebs.io/bdc-protection"
 )
 
 // Add creates a new BlockDeviceClaim Controller and adds it to the Manager. The Manager will set fields on the Controller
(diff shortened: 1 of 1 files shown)


pkg/controller/blockdeviceclaim/blockdeviceclaim_controller.go
```diff
@@ -193,7 +193,11 @@ func (r *ReconcileBlockDeviceClaim) FinalizerHandling(
 		return nil
 	}
 	// The object is being deleted
-	if util.Contains(instance.ObjectMeta.Finalizers, BlockDeviceClaimFinalizer) {
+	// Check if the BDC has only NDM finalizer present. If yes, it means that the BDC
+	// was deleted by the owner itself, and NDM can proceed with releasing the BD and
+	// removing the NDM finalizer
+	if len(instance.ObjectMeta.Finalizers) == 1 &&
+		util.Contains(instance.ObjectMeta.Finalizers, BlockDeviceClaimFinalizer) {
```
kmova: Will this impact BDC's that were created with the older finalizer string in 1.0? Should we as part of upgrade patch all the BDCs with this name change or handle the backward compatibility in the code?

akhilerm: Yes @kmova . We will  need to patch the finalizer as part of upgrade script. I have included this as a part of NDM upgrade steps.

kmova: Also, one more query - in case of manual BDC creation by user. Can there be a case of nil failure if there are no finalizers.

akhilerm: If there are no finalizers, then the `ObjectMeta.Finalizer` array will be empty. Therefore, 
`len(instance.ObjectMeta.Finalizers)` will be equal to zero.

PR #286, opened 2019-07-22
feat(upgrade): perform upgrade from NDM operator

akhilerm:
This PR adds a new pkg to perform upgrade operations. Preupgrade -> upgrade -> postupgrade tasks all can be performed using the package. It is also possible to handle upgrade for multiple release versions.

Currently, only the preupgrade script is being executed. It handles renaming finalizer in BDC.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -110,6 +110,14 @@ func main() {
 		os.Exit(1)
 	}
 
+	log.Info("Registering Components")
+
+	// Setup Scheme for all resources
+	if err := apis.AddToScheme(mgr.GetScheme()); err != nil {
+		log.Error(err, "")
+		os.Exit(1)
+	}
+
 	// Upgrade the components if required
 	k8sClient, err := client.New(cfg, client.Options{})
 	if err != nil {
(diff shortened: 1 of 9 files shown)


cmd/manager/main.go
```diff
@@ -21,8 +21,11 @@ import (
 	"flag"
 	"fmt"
 	"github.com/openebs/node-disk-manager/pkg/setup"
+	"github.com/openebs/node-disk-manager/pkg/upgrade"
+	"github.com/openebs/node-disk-manager/pkg/upgrade/preupgrade"
```
kmova: Is there a need for a preupgrade package? Can it be part of a step under the upgrade itself?

cmd/manager/main.go
```diff
@@ -115,6 +118,20 @@ func main() {
...
 	}
 
+	// Upgrade the components if required
+	k8sClient, err := client.New(cfg, client.Options{})
+	if err != nil {
+		log.Error(err, "")
+		os.Exit(1)
+	}
+
+	log.Info("Performing upgrade operation")
```
kmova: The message can be something like `Check if any of the CRs have to be upgraded`.

cmd/manager/main.go
akhilerm: Actually, my aim is to create a package for each release versions. like `1.0 -> 1.1`. Since only preupgrade is available for the current release, structured like. I shall move it to upgrade pkg itself.

PR #290, opened 2019-07-25
fix(daemon, controller): change nodeName to hostname in disk and BD

akhilerm:
earlier node name was used as `kubernetes.io/hostname`, in disk and blockdevices. But in some cloud providers like AWS, node name and hostname are not same. Therefore, now we fetch the nodename using downward API, and then get the hostname label set on that node.

### DIFF
--- a/cmd/ndm_daemonset/controller/controller.go
+++ b/cmd/ndm_daemonset/controller/controller.go
@@ -187,8 +187,8 @@ func (c *Controller) newClientSet() (client.Client, error) {
 	return clientSet, nil
 }
 
-// getNodeName set HostName field in Controller struct
-// if it gets from env else it returns error
+// setHostName set HostName field in Controller struct
+// from the labels in node object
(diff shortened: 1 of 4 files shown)


cmd/ndm_daemonset/controller/controller.go
```diff
@@ -183,17 +187,34 @@ func (c *Controller) newClientSet() (client.Client, error) {
...
+	}
+	// get the node object and fetch the hostname label from the
+	// node object
+	node := &v1.Node{}
+	err = c.Clientset.Get(context.TODO(), client.ObjectKey{Namespace: "", Name: nodeName}, node)
+	if err != nil {
+		return err
 	}
-	c.HostName = host
+	c.HostName = node.Labels[NDMHostKey]
```
kmova: n_: It is expected that this label will be always persent, but not guaranteed.. let us put an empty string check and fallback to nodename or error out here.

PR #293, opened 2019-08-08
fix(cleaner): add service account to cleanup job

akhilerm:
add service account to cleanup jobs, so that job pods requiring privileged access can be started in openshift cluster.

Fixes [openebs#2698](https://github.com/openebs/openebs/issues/2698)

### DIFF
--- a/pkg/cleaner/jobcontroller.go
+++ b/pkg/cleaner/jobcontroller.go
@@ -95,6 +95,7 @@ func NewCleanupJob(bd *v1alpha1.BlockDevice, volMode VolumeMode, namespace strin
 		podSpec.Volumes = []v1.Volume{volume}
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/cleaner/jobcontroller.go
```diff
@@ -95,6 +95,7 @@ func NewCleanupJob(bd *v1alpha1.BlockDevice, volMode VolumeMode, namespace strin
 		podSpec.Volumes = []v1.Volume{volume}
 	}
 
+	podSpec.ServiceAccountName = getServiceAccount()
```
kmova: What happens if the serviceAccount name ENV is missing will the pod creation fail or use the default service account?

akhilerm: if ENV is missing, we will get a `""` for the service account. Which is equal to not setting any service account.

PR #295, opened 2019-08-15
fix(cleaner): cancel jobs of inactive blockdevices

akhilerm:
This fix makes sure that cleanup jobs are started only on Active BDs. Also, if the BD goes into another state, during cleanup, the job will be cancelled and BD state will remain unchanged. This ensures that stale jobs are not present in the cluster, when corresponding BDs have been removed.

### DIFF
--- a/docs/cleanup-design.md
+++ b/docs/cleanup-design.md
@@ -1,0 +1,40 @@
+# Clean-up of released BlockDevices
+
+NDM Operator performs the cleanup of the data on the blockdevice(BD) before it can be claimed by 
+another BlockDeviceClaim(BDC).
+
+The cleanup operation can be of two types depending on the VolumeMode of the BDC. VolumeMode can be
+- Block : a `wipefs` command will be issued on the BD
+- FileSystem : an `rm -rf` command is issued on the mountpoint of the BD
+
+The cleanup is performed by a kubernetes job, that is scheduled to run on a specified node. The
+following cycle of operations is performed for scheduling a cleanup job.
+
+```
++-----------------+         +------------------+         +------------------+         +---------------+             +------------+
+|                 |         |  Check if status |         |  Check if job is |         | Check status  | Completed   |            |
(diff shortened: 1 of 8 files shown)


docs/cleanup-design.md
```diff
@@ -0,0 +1,40 @@
...
+                                     |                             |                    +------+-----+                     |	
+                                     |                             |                    |            |                     |	
+                                     |                             | Yes                | start job  |                     |	
+                                     v                             |                    |            |                     |	
+                              +------+---------+                   |                    |            |                     |	
+                              |                |                   |                    +-------+----+                     |	
+                              |    return      |<------------------+----------------------------|--------------------------+ 	
+                              +----------------+                                    		  			 
+                                                                                                            	                 
+```
```
kmova: Does a cancel job delete the job or mark it as completed? What happens in the following scenario:
- BD is released. 
- Cleanup Job is scheduled
- During the Cleanup, BD goes inactive - due to node failure or other errors
- ndm-operator sees that BD is inactive and cancels the job. 
--
- After a while the node is back online
- ndm-operator checks the status of the BD is released. and looks for a job. Finds a job that is cancelled/completed? Will the BD remain in released state?

akhilerm: @kmova  A cancel will delete the job, without checking its status (can be completed/running/pending etc).

Workflow for the mentioned sequence of events:
1. BD is released
2. Cleanup Job is scheduled
3. During cleanup, BD goes inactive - due to some reasons
4. NDM-Operator sees that BD is inactive and cancels the job (Delete the job without checking status). The BD status will not be updated.  (will be `Released` itself)
5. In the next reconcile loop of the BD, operator finds that the BD is in released state and no jobs are present for the BD. But it will not start the job since the BD is inactive.

- Step 5 will happen in case of all subsequent reconcile loops till BD becomes `Active`

6. After sometime, node is back online and BD is `Active`
7. In the next reconcile loop, Operator schedules the cleanup job again, since the BD is in `Released` state and is `Active`
8. Once the cleanup job is executed successfully (i.e job marked completed by kubernetes), the job is deleted and BD is marked as `Unclaimed`

PR #296, opened 2019-08-22
feat(apis, controller, manager): include node attributes to disk/BD

akhilerm:
Currently only hostname was included as part of the device information. This has been refactored to include nodename and hostname along with provision for multiple other fields which can be used for geolocating a node and its associated disks and BDs.

**Note for Reviewer**: Most of the test cases in daemonset controller and probes has been refactored in this PR with labelling changes.

### DIFF
--- a/pkg/apis/openebs/v1alpha1/blockdevice_types.go
+++ b/pkg/apis/openebs/v1alpha1/blockdevice_types.go
@@ -29,6 +29,8 @@ import (
 
 // DeviceSpec defines the desired state of BlockDevice
 type DeviceSpec struct {
+	// NodeAttributes has the details of the node in which this BD is attached
+	NodeAttributes  NodeAttribute       `json:"nodeAttributes"`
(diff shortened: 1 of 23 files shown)


kmova:
Looks good @akhilerm - can the sample YAMLs be also updated with the new Spec. Also are there any challenges with upgrades?

pkg/apis/openebs/v1alpha1/blockdevice_types.go
```diff
@@ -29,6 +29,8 @@ import (
 
 // DeviceSpec defines the desired state of BlockDevice
 type DeviceSpec struct {
+	// NodeAttributes has the details of the node in which this BD is attached
+	NodeAttributes  NodeAttribute       `json:"nodeAttributes"`
```
kmova: Where do we say - omit if empty?

akhilerm: `nodeAttributes.nodeName` should always be there in the spec. This why no `omitempty` tag is given

PR #298, opened 2019-08-23
feat(apis, selector): add node name filter

akhilerm:
Added node name filter to BDC.
Use case:
1. If hostname is only given, matching BDs on the host will be used for selection
2. If nodename is only given, matching BDs on the node will be used for selection
3. If both hostname and nodename are provided, BDs with matching hostname and nodename will be used for selection
4. If no node selection value is provided, all BDs will be used for selection

### DIFF
--- a/deploy/crds/openebs_v1alpha1_blockdeviceclaim_cr.yaml
+++ b/deploy/crds/openebs_v1alpha1_blockdeviceclaim_cr.yaml
@@ -4,6 +4,8 @@ metadata:
   name: example-blockdeviceclaim
 spec:
   hostName: "" # hostname of the node from which you want to get a BD
+  nodeAttributes:
+    nodeName: "" # node name of the k8s node. Output from `kubectl get nodes`
   blockDeviceName: "" # BD name, if you want to claim a specific block device
   resources:
     requests:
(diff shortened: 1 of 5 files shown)


deploy/crds/openebs_v1alpha1_blockdeviceclaim_cr.yaml
```diff
@@ -4,6 +4,8 @@ metadata:
   name: example-blockdeviceclaim
 spec:
   hostName: "" # hostname of the node from which you want to get a BD
+  nodeAttributes:
+    nodeName: "" # node name of the k8s node. Output from `kubectl get nodes`
```
kmova: can we also include the `hostname` here. It will be a duplicate, but we can probably deprecate the hostName from above for newer versions. Btw, is spec.hostName optional?

akhilerm: we can include it. Currently the same `NodeAttribute` struct is used for both BD and BDC. BD does not need a `hostname` in its node-attributes as it is present in the label. 
So, will make a new type for the BDC-Node attributes, it will help in deprecating the `hostname` field.

Also, with this PR hostname is optional.

PR #299, opened 2019-08-26
fix(cleaner): fix missing label from cleanup job

akhilerm:
This PR fixes the hostname label on the cleanup job which was missed after the nodename refactoring in #296 

### DIFF
--- a/pkg/cleaner/jobcontroller.go
+++ b/pkg/cleaner/jobcontroller.go
@@ -104,8 +104,8 @@ func NewCleanupJob(bd *v1alpha1.BlockDevice, volMode VolumeMode, namespace strin
 	podTemplate.Spec = podSpec
 
 	labels := map[string]string{
-		controller.KubernetesLabelPrefix: nodeName,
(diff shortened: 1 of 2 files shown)


pkg/cleaner/jobcontroller.go
```diff
@@ -104,8 +104,8 @@ func NewCleanupJob(bd *v1alpha1.BlockDevice, volMode VolumeMode, namespace strin
 	podTemplate.Spec = podSpec
 
 	labels := map[string]string{
-		controller.KubernetesLabelPrefix: nodeName,
-		BDLabel:                          bd.Name,
+		controller.KubernetesHostNameLabel: nodeName,
```
kmova: hmm. should we delete the old label? To catch if there are other places where we might be using it?

akhilerm: We are not deleting the old label. Both `kubernetes.io/hostname` and `blockdevice` label will be present on the job

PR #303, opened 2019-08-29
feat(upgrade): add pre-upgrade tasks for 0.4.1 to 0.4.2

akhilerm:
added preupgrade steps to copy `.spec.hostName` to `.spec.nodeAttributes.hostName`


cmd/manager/main.go
```diff
@@ -148,7 +150,7 @@ func main() {
 
 // performUpgrade performs the upgrade operations
 func performUpgrade(client client.Client) error {
-	//TODO: this task should be named for release, not for upgrade steps
-	preUpgradeTask := upgrade.NewPreUpgradeTask("1.0", "1.1", client)
-	return upgrade.RunUpgrade(preUpgradeTask)
+	v040_v041UpgradeTask := v040_041.NewUpgradeTask("0.4.0", "0.4.1", client)
```
kmova: the package names could be make generic or just include the `toVersion041`, `toVersion042` , the ndm pods might be getting upgrade directly from 040 to 042 as well.

akhilerm: @kmova , But operator will be always doing an incremental upgrade.

kmova: Will take this for now, but we will need to make this into a more generic solution as we add more releases.

PR #304, opened 2019-09-03
fix(filter): fix os-filter to ignore empty exclude paths

akhilerm:
If `os-disk-filter` failed to find the device in which os is installed, the os-filter should not match with any devices. But currently if the filter failed to initialize, all devices on that node were getting excluded. A change has been introduced in the regex to fix this bug. 

Related Issues: [#2727](https://github.com/openebs/openebs/issues/2727)

### DIFF
--- a/cmd/ndm_daemonset/filter/osdiskexcludefilter_test.go
+++ b/cmd/ndm_daemonset/filter/osdiskexcludefilter_test.go
@@ -118,6 +118,11 @@ func TestOsDiskExcludeFilterExclude(t *testing.T) {
 			disk:     &controller.DiskInfo{Path: "/dev/vg0-lv0"},
 			expected: false,
 		},
+		"exclude path is empty and device path is /dev/sda": {
+			filter:   oSDiskExcludeFilter{excludeDevPath: ""},
+			disk:     &controller.DiskInfo{Path: "/dev/sda"},
+			expected: true,
+		},
 	}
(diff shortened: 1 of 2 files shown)


cmd/ndm_daemonset/filter/osdiskexcludefilter_test.go
```diff
@@ -118,6 +118,11 @@ func TestOsDiskExcludeFilterExclude(t *testing.T) {
 			disk:     &controller.DiskInfo{Path: "/dev/vg0-lv0"},
 			expected: false,
 		},
+		"exclude path is empty and device path is /dev/sda": {
+			filter:   oSDiskExcludeFilter{excludeDevPath: ""},
+			disk:     &controller.DiskInfo{Path: "/dev/sda"},
+			expected: true,
```
kmova: What does `true` mean? The disk will be excluded or not excluded?

akhilerm: true means filter passed, and device will not be excluded.

kmova: should we rename `expected` to -- `diskExcluded` ?

kmova: the comments on the function have been updated.

PR #306, opened 2019-09-17
[WIP] Adding builds for ppc64le

Pensu:
Hi,

This is first in a series of PR to enable OpenEBS for ppc64le arch. I have marked it as WIP as some changes require discussion.

1) I have disabled `shellcheck` for ppc64le. It can be added back once it is supported for ppc64le.
2) Added `linux-ppc64le` in .travis.yml to enable builds for ppc64le arch on travis.
3) minikube doesn't provide binaries for any arch other than amd64. Hence, I have disabled some of the steps for ppc64le that could have broken the build before running the actual script i.e. the make command. Is it possible to use `kubeadm` instead of `minikube` as the former one already has multi-arch support?
4) If it is decided to not use `kubeadm`, would it be possible to run tests without `minikube`?
5) build script now builds for native archs, as cross building would require cgo cross dependencies for to be installed on the host.

Please let me know if I need to open an issue to discuss above points? Thanks.

cc: @kmova @harshvkarn

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 dist: xenial
 language: go
 
+os:
+  - linux
+  - linux-ppc64le
+
 go:
   - 1.12.7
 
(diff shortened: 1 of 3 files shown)


kmova:
Thanks @Pensu - (1) and (2) seem ok. 

For (3) and (4),  let us raise another PR to shift to kubeadm or kind or microk8s any one that support multiple arch. 

For (5) we need to setup the multiarch builds. I am exploring using gox. any other suggestions?

Pensu:
@kmova Yeah, we probably would have to move to kubeadm itself, as other options don't have supports for multi-arch. (kind has ppc64le binaries, but it's not fully tested, so there might be consistency issues at later stages.). I will prepare and send a PR to replace minikube with kubeadm.

For cross-builds, we need to make sure the relevant toolchains are available to build with the cross-compiler. Another option to explore could be: https://github.com/karalabe/xgo

Pensu:
Closing in favor of #374

PR #308, opened 2019-09-18
refactor(build): for multiple arch compilation using gox

kmova:
Related to:
- https://github.com/openebs/openebs/issues/1295
- https://github.com/openebs/node-disk-manager/pull/306

This PR refactors the build scripts ( .travis.yaml and Makefile) to allow
for compiling on different platforms like `amd64` and `arm64` without 
having to make changes to the build scripts. 

The following changes have been made: 
- `shellcheck` is triggered as a seperate step from .travis.yaml, as it is only available for amd64
- updated the Makefile to initialize ARCH and BASEIMAGE depending on the platform where builds are initiated. 
- updated the build to make use of gox to allow for cross compilation
- made some cosmetic changes to Makefile for better readability. 
- updated the README with dependency on SeaChest libraries. 

Attempted to automate the building of arm64 builds via cross compiling 
in travis. However, the dependencies on libudev and seachest are non-trivial
to resolve. Will re-attempt to setup automated builds via QEMU based approach
in the upcoming PRs. 

Note: Verified the changes on ARM by manually running the - `make` in a ARM node. 

Signed-off-by: kmova <kiran.mova@mayadata.io>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,13 +6,9 @@ go:
 
 env:
   global:
-    - TAG="ci"
-    - CODECOV_TOKEN="d1e39bea-f800-45de-8266-3e9b85e2594a"
-  matrix:
-    - XC_ARCH="amd64"
-      BASEIMAGE="ubuntu:16.04"
-    #- XC_ARCH="arm64"
-    #  BASEIMAGE="arm64v8/ubuntu:18.04"
+   - ARCH=$(go env GOARCH)
+   - TAG="ci"
+   - CODECOV_TOKEN="d1e39bea-f800-45de-8266-3e9b85e2594a"
 
 sudo: required
 
@@ -22,8 +18,6 @@ services:
 addons:
   apt:
     update: true
-    packages: 
-      - gcc-multilib
 
 install:
   - make bootstrap
@@ -33,11 +27,9 @@ install:
       mv $TRAVIS_BUILD_DIR $GOPATH/src/github.com/openebs;
       cd $GOPATH/src/github.com/openebs/node-disk-manager;
     fi
-  - if [ "$XC_ARCH" == "amd64" ]; then
-      curl -Lo minikube https://storage.googleapis.com/minikube/releases/v1.0.0/minikube-linux-amd64;
-      sudo chmod +x minikube;
-      sudo mv minikube /usr/local/bin/;
-    fi
+  - curl -Lo minikube https://storage.googleapis.com/minikube/releases/v1.0.0/minikube-linux-amd64
+  - sudo chmod +x minikube
+  - sudo mv minikube /usr/local/bin/
   # we need openSeaChest repo to build node-disk-manager
   - pushd .
   - cd ..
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -44,24 +52,26 @@ install:
...
 before_script:
-  - sudo minikube start --vm-driver=none --cpus=1 --memory=1024
-  - sudo minikube update-context
-  # By default root is the owner of the .minikube directory. This is changed to $USER so that
-  # from the integration tests which run as a non-root user can access the files and connect
-  # to the minikube cluster
-  - sudo chown -R $USER $HOME/.minikube
-  - sudo chown -R $USER $HOME/.kube
+  - if [ "$XC_ARCH" == "amd64" ]; then
+      sudo minikube start --vm-driver=none --cpus=1 --memory=1024;
```
akhilerm: the `--cpus` and `--memory` flag can be removed. If VM driver is none, these flags are not considered.

.travis.yml
```diff
@@ -6,9 +6,13 @@ go:
...
   global:
-   - ARCH=$(go env GOARCH)
-   - TAG="ci"
-   - CODECOV_TOKEN="d1e39bea-f800-45de-8266-3e9b85e2594a"
+    - TAG="ci"
+    - CODECOV_TOKEN="d1e39bea-f800-45de-8266-3e9b85e2594a"
+  matrix:
+    - XC_ARCH="amd64"
+      BASEIMAGE="ubuntu:16.04"
+    #- XC_ARCH="arm64"
```
akhilerm: If another `arch` is added to the matrix, will a binary be generated for that also.?

.travis.yml
```diff
@@ -44,24 +52,26 @@ install:
...
   # a particular scsi device, sudo or root permissions are required.
-  - sudo -E env "PATH=$PATH" make test
-  - make integration-test
+  - if [ "$XC_ARCH" == "amd64" ]; then
+      sudo -E env "PATH=$PATH" make test;
+      make integration-test;
+      bash <(curl -s https://codecov.io/bash);
+    fi
 
 after_success:
```
akhilerm: `after_success` should be called only after `script` in all jobs passes right? Otherwise if the testing fails in `amd64`, still image can be pushed for `arm64`

Makefile
```diff
@@ -8,32 +8,44 @@ BUILD_PATH_NDM=ndm_daemonset
...
-XC_OS?= $(shell go env GOOS)
-XC_ARCH?= $(shell go env GOARCH)
-ARCH:=${XC_OS}_${XC_ARCH}
-
 # VERSION is the version of the binary.
 VERSION:=$(shell git describe --tags --always)
 
+ifeq (${XC_OS}, )
+  XC_OS:=$(shell go env GOOS)
+  export XC_OS
```
akhilerm: After `XC_OS` why we need to export it again?

Makefile
```diff
@@ -8,32 +8,44 @@ BUILD_PATH_NDM=ndm_daemonset
...
+endif
+
+ifeq (${XC_ARCH}, )
+  XC_ARCH:=$(shell go env GOARCH)
+  export XC_ARCH
```
akhilerm: same comment as above.

.travis.yml
kmova: this was idea. but reverted this change for now. The Cross compile on travis with CGO dependencies needs to be resolved first.

.travis.yml
kmova: above was mainly about test coverage data being pushed. i reverted this change as well.

Makefile
kmova: These values need to be available from within the build.sh

Makefile
```diff
@@ -1,120 +1,181 @@
...
+endif
+endif
+export BASEIMAGE
 
-# IMAGE is the image name of the node-disk-manager docker image.
+
+# Initialize the NDM DaemonSet variables
+# Specify the binary name
+NODE_DISK_MANAGER=ndm
+NODE_DISK_MANAGER?=ndm
```
akhilerm: We won't be needing `?=` as we are setting the variable on the previous line

Makefile
```diff
@@ -1,120 +1,181 @@
...
+# Specify the path for the main under the ./cmd/
+BUILD_PATH_NDM=ndm_daemonset
+BUILD_PATH_NDM?=ndm_daemonset
```
akhilerm: same comment as above.

Wangzihao18:
> LGTM.
> 
> Just one question: Do we have any specific reason as to use Ubuntu:18.04 for arm and Ubuntu 16.04 for amd64? @kmova

Because Kunpeng920 which is ARM architecture needs to be supported by Ubuntu:18.04.

PR #317, opened 2019-10-06
Continous Integration should run Lint, Flow for every PR

Apollo9999:
The package.json defines a check-all script that runs Prettier, Lint, and Flow. This script is intended to be run before commit to help catch inconsistent formatting or type errors. However we are not currently running this check as part of CI.

  To fix this, first create a script that does the following 
Run Prettier and fail if any modifications are made.This means that the committed code isn't formatted correctly.
Run ESLint and fail if there are any errors.
Run Flow and fail if there are any errors.
  Fix all pre-existing errors so the new yarn ci-check task passes in master.
  Hook this new script into CircleCI to run automatically for every PR

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,7 +54,6 @@ script:
   # a particular scsi device, sudo or root permissions are required.
   - sudo -E env "PATH=$PATH" make test
   - make integration-test
-  - kubectl describe pod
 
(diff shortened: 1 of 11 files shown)


kmova:
This PR doesn't contain the mentioned changes.

PR #318, opened 2019-10-07
feat(ci): migrate minikube to kubeadm via `kind`

### LINKED ISSUE #314: Move from minikube to kubeadm for integration-test
Currently node-disk-manager uses minikube for running integration tests. But while going for cross-platform support, minikube cannot be used for testing. Therefore, the testing needs to be moved from minikube to `kubeadm` based local cluster.

The changes need to be made in [.travis.yml](https://github.com/openebs/node-disk-manager/blob/ebc261b38dc3821663de2175ae48cc94d567f026/.travis.yml#L30). Setting up and starting of minikube,  should be replaced and a cluster using `kubeadm` should be brought up

distributedlock:


RanadeepPolavarapu:
I got `kind` (kubernetes-in-docker) to bootstrap via `kubeadm`. Could someone help with why integration tests are failing?

akhilerm:
If you see the Travis log,you can see that build failed. not the actual tests. Can you check whether `gox` is available on the VM?

.travis.yml
```diff
@@ -23,38 +23,43 @@ install:
...
+  - export KUBECONFIG="$(kind get kubeconfig-path)"
+  - echo $KUBECONFIG
+  - ln -s $KUBECONFIG $HOME/.kube/config
+  - ls -latr $HOME/.kube
+  - kind get clusters
 
 script:
   - make shellcheck
-  - make
+  - sudo make
```
akhilerm: Why is `sudo` required here?

akhilerm (commented):
When using `kind` are the devices on the host  available inside the cluster?

akhilerm:
@RanadeepPolavarapu  I tried setting up a single node cluster using kind. I was not able to detect the disk attach and detach that we do as part of integration testing. Can this be changed to just `kubeadm` based local cluster instead of using `kind` . This will make the travis VM itself the node in the cluster.

RanadeepPolavarapu:
I tried to make it just kubeadm initially but it was having issues bootstrapping the cluster due to Travis itself being run in a container. I did some research there was a method to do kubeadm in docker for CI environments that people were using two years ago. Google recently added kind and deprecated that method of kubeadm in docker since it was a bit more hacky. Kind essentially does the same thing with kubeadm bootstrapping the cluster but optimized for inside a container. 

I'll spend more time on this on the weekend.

.travis.yml
RanadeepPolavarapu: The `make` was performing a clean of the *go* directory.

https://github.com/openebs/node-disk-manager/blob/a63af3e2e965a95e36c1f2d4d8e567d79e0f3200/.travis.yml#L33

This was adding pkgs as root or something. I'll debug and if needed I'll `chown` it back to the current user (travis) `$(id)`.

akhilerm:
@RanadeepPolavarapu  Are there any updates on the changes requested?

akhilerm:
@RanadeepPolavarapu  did you make the requested changes?

akhilerm:
Can this be done in a similar way how `local-cluster` is done in kubernetes?

kmova:
This PR has been inactive for a long time. Please re-open if you have new changes to push into this PR.

PR #333, opened 2019-11-06
Fixes device detection for KVM-based virtualization

### LINKED ISSUE #188: NDM tries to reuse the same name for disks on different hosts (and fails)
**What happened**:

I have three Proxmox nodes, each with a 100GB disk attached to `/dev/sdb`. Only one disk shows up under `kubectl get disk`, and the logs indicate that each NDM pod is attempting to add its own disk with the same name and failing because one already exists with that name. My best guess is that it's hashing the disk name, and since they're all attached as `scsi-0QEMU_QEMU_HARDDISK_drive-scsi1` under `/dev/disk/by-id`, the hash returns the same value.

**What you expected to happen**:

I expected to have all three disks appear.

**How to reproduce it (as minimally and precisely as possible)**:

- spin up a three-node Proxmox cluster
- attach your Synology NAS to it
- build a Rancher/RKE installation onto three VMs
- attach three identical iSCSI/LVM disks to the rancher nodes
- install OpenEBS
- see that only one disk shows up
- check the logs and see that two of the pods show:

```
I1229 23:52:57.633681       1 diskstore.go:40] create failed, trying to update the disk object : disks.openebs.io "disk-51e5c068c21f5186044957960dfbaefb" already exists
```

**Anything else we need to know?**:

I was able to work around this by creating the Disk resources manually with names like `disk-sdb-rancher01`.

footplus:
This PR aims to resolve #188 for KVM-based virtualization, while awaiting global changes on disk detection strategy (https://github.com/openebs/openebs/pull/2666)

Signed-off-by: Aurelien Guillaume <aurelien@iwi.me>

### DIFF
--- a/pkg/udev/common.go
+++ b/pkg/udev/common.go
@@ -121,6 +121,7 @@ func (device *UdevDevice) GetUid() string {
 	localDiskModels := make([]string, 0)
 	localDiskModels = append(localDiskModels, "EphemeralDisk")
 	localDiskModels = append(localDiskModels, "Virtual_disk")
+	localDiskModels = append(localDiskModels, "QEMU_HARDDISK")
 	if len(idtype) == 0 || util.Contains(localDiskModels, model) {
 		// as hostNetwork is true, os.Hostname will give you the node's Hostname
 		host, _ := os.Hostname()


pkg/udev/common.go
```diff
@@ -121,6 +121,7 @@ func (device *UdevDevice) GetUid() string {
 	localDiskModels := make([]string, 0)
 	localDiskModels = append(localDiskModels, "EphemeralDisk")
 	localDiskModels = append(localDiskModels, "Virtual_disk")
+	localDiskModels = append(localDiskModels, "QEMU_HARDDISK")
```
akhilerm: This will help to identify KVM disks when the disk is not having a unique serial number right?

footplus: Exactly.

footplus:
Yes, exactly. In fact a more tunable way to configure the disk detection is required, but this is managed by openebs/openebs#2666 ; the goal here is to permit working on clouds which are exposing QEMU_HARDDISK volumes, at specific /dev/sdX device names, but which bears strictly identical characteristics otherwise (like Virtual_Disk and others, it's the exact same use-case).

Currently, it's impossible to use ndm if you have such disks, because all disks bear the same UUID.

PR #349, opened 2019-11-25
 feat(exporter): add collector for fetching metrics using seachest library

akhilerm:
- added node level exporter for fetching metrics using seachest library. 
- In this PR, only the current temperature info of the disk is exposed as metrics. More metrics like disk rotation rate can be fetched from seachest library. 
- More collectors can be added by following steps in #347 .

### DIFF
--- a/blockdevice/blockdevice.go
+++ b/blockdevice/blockdevice.go
@@ -114,6 +114,9 @@ type BlockDevice struct {
 	// eg: dm-0 is a slave to sda1. Then the list of dm-0 will contain sda1
 	Slaves []string
 
+	// TemperatureInfo stores the temperature information of the drive
+	TemperatureInfo TemperatureInformation
+
 	// Status contains the state of the blockdevice
 	Status Status
 }
@@ -154,6 +157,16 @@ type FileSystemInformation struct {
 
 	// MountPoint is the list of mountpoints at which this blockdevice is mounted
 	MountPoint []string
+}
+
+// TemperatureInformation stores the temperature information of the blockdevice
+type TemperatureInformation struct {
+	// TemperatureDataValid specifies whether the current temperature
+	// data reported is valid or not
+	TemperatureDataValid bool
+
+	// CurrentTemperature is the temperature of the drive in celsius
+	CurrentTemperature int16
 }
 
 // Status is used to represent the status of the blockdevice
--- a/ndm-exporter/collector/seachest.go
+++ b/ndm-exporter/collector/seachest.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2019 The OpenEBS Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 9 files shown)


blockdevice/blockdevice.go
```diff
@@ -156,6 +159,16 @@ type FileSystemInformation struct {
 	MountPoint []string
 }
 
+// TemperatureInformation stores the temperature information of the blockdevice
+type TemperatureInformation struct {
+	// TemperatureDataValid specifies whether the current temperature
+	// data reported is valid or not
```
kmova: how do you determine if the temperature value is valid?

ndm-exporter/collector/seachest.go
```diff
@@ -0,0 +1,180 @@
...
+	"github.com/openebs/node-disk-manager/db/kubernetes"
+	seachestmetrics "github.com/openebs/node-disk-manager/pkg/metrics/seachest"
+	"github.com/openebs/node-disk-manager/pkg/seachest"
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/klog"
+)
+
+// SeachestMetricCollector contains the metrics, concurrency handler and client to get the
+// metrics from seachest
+type SeachestMetricCollector struct {
```
kmova: Should we keep this generic. For example this is SMART metrics that we are getting and in the current implementation we are using SeaChest to get the values.

blockdevice/blockdevice.go
akhilerm: Seachest gives that information. For every temperature information that seachest provides there is an associated boolean value, which specifies whether the data is valid or not

ndm-exporter/collector/seachest.go
akhilerm: Its more towards the library we use. We can use seachest library for some metrics. SMART for some other metrics. There will be some metrics that overlaps. But I think a library specific metric will be good, even if they overlaps. This will give the option to the user on which metric he need to choose.

If we think in angle of node exporter, there will be some disk related metrics that will be exposed by both node-exporter and NDM. But the user can choose the metric using his query, rather than we limiting the user to metrics from a certain lib only.

ndm-exporter/collector/seachest.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+// collectErrors collects only the error metrics and set it on the channel
+func (sc *SeachestCollector) collectErrors(ch chan<- prometheus.Metric) {
+	for _, col := range sc.metrics.ErrorCollectors() {
+		col.Collect(ch)
+	}
+}
+
+// GetMetricData gets the seachest metrics for each blockdevice and fills it in the blockdevice struct
```
kmova: Should this method be exposed outside of the package? Similar check needs to be performed to other method to set the visibility as appropriate.

akhilerm: Done. methods are now not available outside the package

PR #358, opened 2019-12-26
fix(script): fix changing the system core pattern

akhilerm:
Earlier NDM used to rewrite the system core pattern so that the core
dump is available in the persistent `/var/openebs/sparse` directory.
But this will cause all process on the host to write core files to that directory.
Now NDM process will be launched from the openebs directory, so core
files will get automatically written to the `$PWD`. This will help to
remove changing the system core pattern.

### DIFF
--- a/build/ndm-daemonset/entrypoint.sh
+++ b/build/ndm-daemonset/entrypoint.sh
@@ -4,8 +4,10 @@ export GOTRACEBACK=crash
 
 echo "[entrypoint.sh] enabling core dump."
 ulimit -c unlimited
+echo "[entrypoint.sh] creating /var/openebs/sparse if not exists."
+mkdir -p /var/openebs/sparse
 echo "[entrypoint.sh] changing directory to /var/openebs/sparse"
-cd /var/openebs/sparse || exit
+cd /var/openebs/sparse
 echo "[entrypoint.sh] launching ndm process."
 /usr/sbin/ndm start &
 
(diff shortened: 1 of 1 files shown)


build/ndm-daemonset/entrypoint.sh
```diff
@@ -4,7 +4,10 @@ export GOTRACEBACK=crash
 
 echo "[entrypoint.sh] enabling core dump."
 ulimit -c unlimited
-echo "/var/openebs/sparse/core.%e.%p.%h.%t" > /proc/sys/kernel/core_pattern
+echo "[entrypoint.sh] creating /var/openebs/sparse if not exists."
+mkdir -p /var/openebs/sparse
+echo "[entrypoint.sh] changing directory to /var/openebs/sparse"
+cd /var/openebs/sparse || exit
```
pawanpraka1: will it work if SPARSE_FILE_DIR env is different than /var/openebs/sparse? should we use SPARSE_FILE_DIR env instead /var/openebs/sparse?

akhilerm: Only the mountPath is dependent. `SPARSE_FILE_DIR` is the directory inside the container .

pawanpraka1: right, so if I have changed that path to something else what will happen? since in the entrypint.sh it is fixed to /var/openebs/sparse.

akhilerm: if the mountPath is also changed, the core dump wont be persisted.  Will raise a PR to fix it. The `SPARSE_FILE_DIR` and mountPath should always be the same.

PR #362, opened 2020-01-09
feat(script, yaml): add openebs basepath

akhilerm:
OpenEBS basepath has been added. Each component of openebs will have a directory under OPENEBS_IO_BASE_DIR on the host machine. 

NDM components will be stored under `OPENEBS_IO_BASE_DIR/ndm` on the host. Inside the NDM container this path will be mapped to `/var/openebs/ndm`.

### DIFF
--- a/build/ndm-daemonset/entrypoint.sh
+++ b/build/ndm-daemonset/entrypoint.sh
@@ -2,19 +2,28 @@
 
 export GOTRACEBACK=crash
 
+# openebs base directory inside the container
+OPENEBS_BASE_DIR="/var/openebs"
+# ndm base directory inside the container. It will be the ndm
+# directory inside openebs base directory
+NDM_BASE_DIR="${OPENEBS_BASE_DIR}/ndm"
+
 # set ulimit to 0, if the core dump is not enabled
 if [ -z "$ENABLE_COREDUMP" ]; then
   ulimit -c 0
 else
   # making sure mountpath inside the container is available
-  cd /var/openebs/ndm || (echo "OpenEBS Base directory not found" && exit 1)
+  if ! [ -d "${NDM_BASE_DIR}" ]; then
+    echo "OpenEBS/NDM Base directory not found"
+    exit 1
+  fi
   # set ulimit to unlimited and create a core directory for creating coredump
   echo "[entrypoint.sh] enabling core dump."
   ulimit -c unlimited
-  echo "[entrypoint.sh] creating /var/openebs/ndm/core if not exists."
-  mkdir -p "/var/openebs/ndm/core"
-  echo "[entrypoint.sh] changing directory to /var/openebs/ndm/core"
-  cd "/var/openebs/ndm/core" || exit
+  echo "[entrypoint.sh] creating ${NDM_BASE_DIR}/core if not exists."
+  mkdir -p "${NDM_BASE_DIR}/core"
+  echo "[entrypoint.sh] changing directory to ${NDM_BASE_DIR}/core"
+  cd "${NDM_BASE_DIR}/core" || exit
 fi
 
 echo "[entrypoint.sh] launching ndm process."
--- a/build/ndm-daemonset/entrypoint.sh
+++ b/build/ndm-daemonset/entrypoint.sh
@@ -7,7 +7,7 @@ if [ -z "$ENABLE_COREDUMP" ]; then
   ulimit -c 0
 else
   # making sure mountpath inside the container is available
-  cd /var/openebs || (echo "OpenEBS Base directory not found" && exit)
+  cd /var/openebs || (echo "OpenEBS Base directory not found" && exit 1)
   # set ulimit to unlimited and create a core directory for creating coredump
   echo "[entrypoint.sh] enabling core dump."
   ulimit -c unlimited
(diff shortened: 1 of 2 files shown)


ndm-operator.yaml
```diff
@@ -180,6 +178,9 @@ spec:
         hostPath:
           path: /proc
           type: Directory
+      - name: basepath
```
vishnuitta: this has to be env isn't it, so that, helm variable can be added? cc: @prateekpandey14

build/ndm-daemonset/entrypoint.sh
```diff
@@ -9,10 +9,10 @@ else
   # set ulimit to unlimited and create a core directory for creating coredump
   echo "[entrypoint.sh] enabling core dump."
   ulimit -c unlimited
-  echo "[entrypoint.sh] creating $SPARSE_FILE_DIR/core if not exists."
-  mkdir -p "$SPARSE_FILE_DIR/core"
-  echo "[entrypoint.sh] changing directory to $SPARSE_FILE_DIR/core"
-  cd "$SPARSE_FILE_DIR/core" || exit
+  echo "[entrypoint.sh] creating /var/openebs/core if not exists."
+  mkdir -p "/var/openebs/core"
```
mittachaitu: Is it good to do something like
```
cd /var/openebs || exit
mkdir -p core
cd core
```
So that even if users given wrong value in mount paths we will not allow.

ndm-operator.yaml
```diff
@@ -139,6 +139,8 @@ spec:
         - name: procmount
           mountPath: /host/proc
           readOnly: true
+        - name: basepath
+          mountPath: /var/openebs
         - name: sparsepath
```
vishnuitta: is this going to impact the path of the file on blockDevice CR?

ndm-operator.yaml
```diff
@@ -180,6 +178,9 @@ spec:
         hostPath:
           path: /proc
           type: Directory
+      - name: basepath
+        hostPath:
+          path: /var/openebs/ndm
```
pawanpraka1: add:- type: DirectoryOrCreate

ndm-operator.yaml
```diff
@@ -180,6 +178,9 @@ spec:
...
       - name: sparsepath
         hostPath:
           path: /var/openebs/sparse
```
pawanpraka1: here also add `type: DirectoryOrCreate`

ndm-operator.yaml
```diff
@@ -180,6 +178,9 @@ spec:
         hostPath:
           path: /proc
           type: Directory
+      - name: basepath
+        hostPath:
+          path: /var/openebs/ndm
       - name: sparsepath
```
pawanpraka1: we need to change helm also, right? We should remove sparsepath and add basepath in the helm?

ndm-operator.yaml
prateekpandey14: ENV not required, we can just simply read from the `values.yaml` file.
We have to make sure that, we should not be having multiple naming paths, to avoid the confusion.

ndm-operator.yaml
akhilerm: Reverting this change.

ndm-operator.yaml
akhilerm: Done.

build/ndm-daemonset/entrypoint.sh
akhilerm: Done.

ndm-operator.yaml
akhilerm: The changes to helm will be done in [this](https://github.com/openebs/openebs/pull/2880) PR in Openebs

build/ndm-daemonset/entrypoint.sh
```diff
@@ -6,13 +6,15 @@ export GOTRACEBACK=crash
 if [ -z "$ENABLE_COREDUMP" ]; then
   ulimit -c 0
 else
+  # making sure mountpath inside the container is available
+  cd /var/openebs || (echo "OpenEBS Base directory not found" && exit)
```
pawanpraka1: here you should use exit 1 (or any non zero value)

akhilerm: Done.

build/ndm-daemonset/entrypoint.sh
```diff
@@ -2,17 +2,28 @@
...
+# directory inside openebs base directory
+NDM_BASE_DIR="${OPENEBS_BASE_DIR}/ndm"
+
 # set ulimit to 0, if the core dump is not enabled
 if [ -z "$ENABLE_COREDUMP" ]; then
   ulimit -c 0
 else
+  # making sure mountpath inside the container is available
+  if ! [ -d "${NDM_BASE_DIR}" ]; then
+    echo "OpenEBS/NDM Base directory not found"
```
mittachaitu: nit: I think to represent the message we should use `$OPENEBS_BASE_DIR/$NDM_BASE_DIR`

akhilerm: `NDM_BASE_DIR=OPENEBS_BASE_DIR/ndm`. That's why used that error message.

build/ndm-daemonset/entrypoint.sh
```diff
@@ -2,17 +2,28 @@
 
 export GOTRACEBACK=crash
 
+# openebs base directory inside the container
+OPENEBS_BASE_DIR="/var/openebs"
```
kmova: Is this variable required?

akhilerm: This variable helps to understand how the NDM base path is constructed using the openebs base directory inside the container. Other than that it provides no additional information and is not used.

PR #365, opened 2020-01-20
chore(vendor): update the dependencies to latest kubernetes version

akhilerm:
update project dependencies
- kubernetes : v1.17.4
- controller-runtime : v0.5.2
- operator-sdk : v0.17.0


kmova:
@akhilerm, will hold this till the github action is added to verify multi-arch builds on pull requests. It would be good to ensure that builds don't fail especially on PRs like this that involve a huge changes.

akhilerm:
@kmova , That means we will have to first prioritize the testing and verification using minikube.

Shall we split into 2 things:
1. We will build multiarch images for each pull request, but will not push it or test it. (`node-disk-<component>-ci` image will be used). This will guarantee that the build works fine.
2. Add integration tests using minikube in github actions.

These 2 will be separate PRs and after `1.` is merged, we should merge this PR,  IMHO

Currently it is not possible to retag images to a different repository using buildx, also mulitarch images cannot stored locally without pushing to an intermediate repository.

As suggested from moby team 
```
... can use intermediate registry or you can just build a single arch for test. after you build multi-arch it will reuse the previous cache.
the advantage of running single arch for tests, is that qemu emulation is slow, so it’s faster to run the test on its native platform.
```

So the integration tests will be run only on one native platform for now.

akhilerm:
@kmova @harshthakur9030 PTAL.

harshthakur9030:
@akhilerm LGTM

akhilerm:
@kmova @harshthakur9030  PTAL. rebased the changes from go version upgrade.

PR #373, opened 2020-02-13
refact(blockdevice): make changes to core blockdevice struct

akhilerm:
- added Identifier field consisting of unique IDs which can be used to identify the blockdevice
- added Capacity field having all storage related information
- added DeviceDetails field consisting of hardcoded information that may be available on the blockdevice
- refactored the exporter code to make use of new fields in BlockDevice


blockdevice/blockdevice.go
```diff
@@ -101,6 +101,9 @@ type BlockDevice struct {
 	// DeviceType is the type of the blockdevice. can be sparse/disk/partition etc
 	DeviceType string
 
+	// DriveType is the type of backing drive for this blockdevice. HDD/SSD/sparse
+	DriveType string
```
shovanmaity: @akhilerm what is the JSON path for DriveType and DeviceType in block device?

akhilerm: This is the `BlockDevice` struct that will be used internally. We are deprecating the `DiskInfo` and `DeviceInfo` struct in favour of this `BlockDevice` struct which will form the core. 

The JSON path is defined for the `BlockDevice` custom resource, `.spec.details.deviceType`.
There is no drive type field currently in `BlockDevice` resource.

db/kubernetes/client.go
```diff
@@ -22,9 +22,10 @@ import (
 	"os"
 	"strings"
 
-	"github.com/openebs/node-disk-manager/blockdevice"
+	. "github.com/openebs/node-disk-manager/blockdevice"
```
kmova: can we continue to use the scope of blockdevice - It helps with the readability of the code.

akhilerm: do you mean, no need to use the dot imports?

akhilerm: Done.

blockdevice/blockdevice.go
```diff
@@ -159,6 +181,81 @@ type FileSystemInformation struct {
 	MountPoint []string
 }
 
+// CapacityInformation holds the capacity related information for the device
+type CapacityInformation struct {
+	// Storage is the storage capacity of this blockdevice
+	// in bytes
+	Storage uint64
+
+	// PhysicalBlockSize is the physical block size in bytes
```
akhilerm: @pawanpraka1 , should the `PhysicalBlockSize`, `LogicalBlockSize` and `HardwareSectorSize` be a part of `CapacityInformation` or `DeviceInformation`

pawanpraka1: it should be part of DeviceInformation as they are the attributes of the device.

blockdevice/blockdevice.go
```diff
@@ -159,6 +181,81 @@ type FileSystemInformation struct {
 	MountPoint []string
 }
 
+// CapacityInformation holds the capacity related information for the device
+type CapacityInformation struct {
```
pawanpraka1: lets rename CapacityInformation to DeviceAttributes, then it makes sense to keep all of them in same struct. DeviceInformation will be used for identifier related information.

akhilerm: We already have a `DeviceInformation` right. Shall we rename that to `DeviceAttributes` and move the blocksize and sector size to that struct. 

I prefer `CapacityInformation` because we need something that says about the actual capacity of the blockdevice. We will keep only `Storage` in `CapacityInformation`

For identifier related information we have an `Identifier` field.

pawanpraka1: seems to be fine.

akhilerm:
```
type BlockDevice struct {
	Identifier
	.
	.
	.
	Capacity CapacityInformation
	DevLinks []DevLink
	DeviceDetails DeviceAttributes
}
type Identifier struct {
	UUID string
	SysPath string
	DevPath string
}
type CapacityInformation struct {
	Storage uint64
}
type DeviceAttributes struct {
	PhysicalBlockSize uint32
	LogicalBlockSize uint32
	HardwareSectorSize uint32
	DeviceType string
	DriveType string
	WWN string
	Vendor string
	Model string
	.
	.
}
```
This is how it will look after the change.

akhilerm: Done the changes as discussed.

blockdevice/blockdevice.go
shovanmaity: @akhilerm do we need to update API struct also?

blockdevice/blockdevice.go
shovanmaity: @akhilerm do we need to update API struct also?

blockdevice/blockdevice.go
akhilerm: This PR contains changes to the internal struct only. Changes to the API and its uses can be updated in further PRs. 
I would prefer it to be done, along with the sysfs probe where logical size, type of drive etc can be updated.

blockdevice/blockdevice.go
shovanmaity: Should `PhysicalSectorSize` and `LogicalSectorSize` part of this?

akhilerm: They are same as `PhysicalBlockSize` and `LogicalBlockSize` given by the system. I have added the comments describing how it is used.

PR #376, opened 2020-02-20
feat(core): deprecate disk resource in favour of blockDevice resource

akhilerm:
Disk resource is deprecated in favour of the BlockDevice resource.

- BlockDevice core struct replaces DiskInfo struct
- WWN is fetched while filling details by udev
- Default UUID prefix changed from `disk` to `blockdevice`


cmd/ndm_daemonset/controller/blockdevice.go
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 package controller
```
kmova: Can this file be completely replaced by `blockdevice/blockdevice.go` ?

cmd/ndm_daemonset/controller/disk.go
```diff
@@ -17,8 +17,8 @@ limitations under the License.
 package controller
```
kmova: Can this file be completely replaced by `blockdevice/blockdevice.go` ?

cmd/ndm_daemonset/controller/blockdevice.go
akhilerm: Not at the moment @kmova . We are generating the `BlockDevice` resource from this struct. Since we haven't written any methods for our new core blockdevice struct to convert to `BlockDevice` resource, we cannot remove this file as of now. 

Once all the client methods are implemented for the new core struct, we can remove this. Those changes will be taken care as part of another PR.

cmd/ndm_daemonset/controller/disk.go
akhilerm: Yes, this file can be removed. But not with `blockdevice/blockdevice.go`. It can be completely removed, as `DiskInfo` is no more used. Even if we want to create a `Disk` resource again at some point, we can generate it from the `BlockDevice` struct

blockdevice/blockdevice.go
```diff
@@ -21,9 +21,12 @@ package blockdevice
 //
 // 1. Example blockdevice struct for a partition /dev/sda1
 // 		{
-// 				UUID:blockdevice-4c25d69f9adc868f61e3d891cf3a5613
+// 				Identifier:{
```
pawanpraka1: this is a breaking change. hoping other dependant components are aware if this.

akhilerm: Its in the internal struct, not in the kubernetes API resource.

PR #378, opened 2020-02-28
feat(core,partition): add partition support to NDM

akhilerm:
add support for partitions in NDM. The PR contains the implementation of the algorithm in [OEP#2666](https://github.com/openebs/openebs/pull/2666)

- add support for partitions when new device is added
- use the new algorithm for creating UUIDs
- add support for detaching partitioned devices
- mount /dev filesystem
- fix probes to fill details which are exclusive to each probe.
- fetch partition number, partition table (UUID/type), partition UUID, filesystem UUID from udev
- added PartitionInfo and FileSystemUUID fields in core BlockDevice struct

**Notes to reviewer**
- Commit b7ef6b8c8a2ca9662e00ee1fae47366d067b52ae is adding vendor dependencies. It can be excluded from reivew. Actual review may start from commit b9cb79f5615878f26b2795d334e6f7e8d1c664d5


cmd/ndm_daemonset/probe/capacityprobe.go
```diff
@@ -92,16 +95,15 @@ func (cp *capacityProbe) FillBlockDeviceDetails(blockDevice *blockdevice.BlockDe
...
-	}
+
+	// The size (/size) entry returns the `nr_sects` field of the block device structure.
+	// Ref: https://elixir.bootlin.com/linux/v4.4/source/fs/block_dev.c#L1267
+	//
+	// Traditionally, in Unix disk size contexts, “sector” or “block” means 512 bytes,
+	// regardless of what the manufacturer of the underlying hardware might call a “sector” or “block”
+	// Ref: https://elixir.bootlin.com/linux/v4.4/source/fs/block_dev.c#L487
+	//
+	// Therefore, to get the capacity of the device it needs to always multiplied with 512
```
pawanpraka1: nice, so this was a bug. On google cloud with local attached SSD, we will see higher size that the disk size attached.

cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -101,3 +153,25 @@ func (pe *ProbeEvent) initOrErrorEvent() {
 		klog.Error(err)
 	}
 }
+
+// generateUUID creates a new UUID based on the algorithm proposed in
+// https://github.com/openebs/openebs/pull/2666
+func generateUUID(bd blockdevice.BlockDevice) (string, bool) {
+	if bd.DeviceAttributes.DeviceType == libudevwrapper.UDEV_PARTITION {
+		klog.Info("device is a partition, using partition UUID:", bd.PartitionInfo.PartitionEntryUUID)
```
pawanpraka1: add devpath also in the log so that we can correlate..

cmd/ndm_daemonset/probe/capacityprobe.go
akhilerm: yes. But we haven't faced the issue because the value that was filled in was over written by smartprobe(because capacity-probe had lowest priority). If smart probe is disabled, the value fetched by capacity probe for local SSDs were wrong.

cmd/ndm_daemonset/probe/eventhandler.go
akhilerm: Done.

cmd/ndm_daemonset/probe/udevprobe.go
```diff
@@ -142,28 +144,49 @@ func (up *udevProbe) scan() error {
...
 				klog.Infof("Dependents of %s : %+v", deviceDetails.DevPath, dependents)
 			}
 		}
 		newUdevice.UdevDeviceUnref()
 	}
-	up.controller.DeactivateStaleBlockDeviceResource(disksUid)
+	if !up.controller.FeatureGates.IsEnabled(features.GPTBasedUUID) {
+		up.controller.DeactivateStaleBlockDeviceResource(disksUid)
+	} else {
+		// TODO deactivate all the stale blockdevices
```
pawanpraka1: we should handle it otherwise removed devices will remain active.

pawanpraka1:
@akhilerm can you merge the changes into two commits, one for vendor and one for the business logic?

pkg/udevevent/event.go
```diff
@@ -43,10 +44,37 @@ func (e *event) process(device *libudevwrapper.UdevDevice) {
...
+
+	// fields used for dependents. dependents cannot be obtained while
+	// removing the device since sysfs entry will be absent
+	if action != libudevwrapper.UDEV_ACTION_REMOVE {
+		devicePath := hierarchy.Device{
+			Path: deviceDetails.DevPath,
+		}
+		dependents, err := devicePath.GetDependents()
+		if err != nil {
+			klog.Errorf("could not get dependents for %s, %v", devicePath, err)
```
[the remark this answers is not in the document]
akhilerm: yes. I think error in getting dependent devices shouldn't be a criteria to halt the event processing. Should we ignore the event if it errors out? WDYT @kmova

cmd/ndm_daemonset/probe/udevprobe.go
akhilerm: yes, we should deactivate them. But when to deactivate? 

Earlier we used to deactivate only those which were not identified. 

Shall we deactivate  all resources on the given host first and then let the event handler update them as active one by one. Are there any corner cases for this approach.?

cmd/ndm_daemonset/probe/capacityprobe.go
kmova: There must be new devices that show the sectorsize as 4K. ( cc: @richardelling ) 

Also, just need to make sure we are not re-opening the issue fixed here: https://github.com/openebs/node-disk-manager/pull/150

kmova:
This PR is being split into smaller PRs.

cmd/ndm_daemonset/probe/capacityprobe.go
akhilerm: New devices(SSDs) show `hw_sector_size` as 4k. But to get the capacity of the device, we need to always multiply the value in `/sys/class/block/sda/size` with 512. 

This is because the `size` file contains the no.of blocks on the block device and it is always calculated by using 512 byte  irrespective of the hardware sector size of the device

cmd/ndm_daemonset/probe/udevprobe.go
akhilerm: @pawanpraka1  WDYT about this approach?

pawanpraka1: hmm, deactivating all of them seems to be ok as if it was graceful shutdown they will be Inactive anyway. I don't see any corner case just that it will be extra etcd call for the devices that are currently attached one for making them Incative and other for making them active. What is the problem in calling DeactivateStaleBlockDeviceResource here? We can have disksUid map created for feature GPTBasedUUID) and pass it here?

akhilerm:
Closing. All the changes in this PR has been split into  multiple smaller PRs #403 , #387 , #386 , #385 , #382  and merged.

PR #382, opened 2020-03-06
feat(core, udev): add partition related information to blockdevice

akhilerm:
- added PartitionInfo struct which contains information if the given blockdevice is a partition. The information includes, partition number, parition UUID, partition table UUID and partition table type.
- added FileSystemUUID which will store the UUID of the filesystem, if a filesystem exists on the blockdevice
- fetch partition number, partition table (UUID/type), partition UUID from udev
- fetch partition number, partition table (UUID/type), partition UUID from udev
- fill all details required for creating UUID during udev event, instead of relying on additional probe. This will help to generate UUID even if all the probe fails.

### DIFF
--- a/blockdevice/blockdevice.go
+++ b/blockdevice/blockdevice.go
@@ -28,6 +28,7 @@ package blockdevice
 //				}
 // 				NodeAttributes:map[hostname:my-machine]
 // 				FSInfo:{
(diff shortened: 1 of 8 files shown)


akhilerm:
This PR contains the first set of changes from #378

PR #383, opened 2020-03-06
refact(seachest,probe): remove unused attributes from seachest probe

akhilerm:
Attributes that are guaranteed to be filled by other probes are removed from seachest. 
seachest-probe will be used to fetch details that are exclusively given by seachest library.

Note: seachest can't be used to fill in storage/capacity related fields,
because the library will always report the capacity/blocksize of the main
disk even if the query is made for the partition. This will result in wrong
information being filled in the blockdevice resource. The sysfs probe will be
used to get all capacity related information from the sysfs filesystem

The following fields which are removed will be filled using:
Controller : Hostname
UdevProbe : DevPath, Model, WWN, Serial, Vendor
SysfsProbe : Storage, DriveType

Some fields like RotationRate, Endurance etc has been removed as they are still not yet supported by the BlockDevice resource.

### DIFF
--- a/cmd/ndm_daemonset/probe/seachestprobe.go
+++ b/cmd/ndm_daemonset/probe/seachestprobe.go
@@ -102,41 +102,6 @@ func (scp *seachestProbe) FillBlockDeviceDetails(blockDevice *blockdevice.BlockD
 		return
 	}
 
-	if blockDevice.DevPath == "" {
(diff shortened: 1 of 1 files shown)


kmova:
@akhilerm -- can you update the PR description with the probes that fill the removed attributes.

akhilerm:
@kmova Updated the PR description with the fields and probes which will be used to fill those details.

PR #386, opened 2020-03-09
feat(partition, uuid): add partition support and UUID implementation

akhilerm:
This PR fixes the issue of disk unique identification in NDM. It implements the algorithm mentioned in [OEP#2666](https://github.com/openebs/openebs/pull/2666). This will make NDM uniquely identify disks without an identifier in a cluster. Disk movement from one node to another, creating block device resources for partitions is also supported.

**What this PR does?**:
- add support for partitions in NDM
- implement the new UUID algorithm mentioned in [#2666](https://github.com/openebs/openebs/pull/2666)
- changes made on the a claimed BlockDevice (like creating partitions) are ignored.
- if partitions are created on an unclaimed device, the device is marked inactive and resources will be created for the partitions.

Yes, But not automated.

1. disk with WWN - single BlockDevice resource is created
2. disk with WWN and 2 partitions - 2 Block Device resources are created for each partition
3. disks of type 1) and 2) are connected and disconnected while NDM is running
4. disks of type 1) and 2) are removed from one node and attached to another node
5. cstor clamining and creating pool on a disk with WWN and a partitioned disk
6. formatting the device with ext4 after claiming and NDM pod is restarted. Filesystem details are not updated on the the resource

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests
- [x] (Optional) If documentation changes are required, which issue on https://github.com/openebs/openebs-docs is used to track them:

### DIFF
--- a/blockdevice/blockdevice.go
+++ b/blockdevice/blockdevice.go
@@ -114,23 +114,8 @@ type BlockDevice struct {
 	// PartitionInfo contains details if this blockdevice is a partition
 	PartitionInfo PartitionInformation
 
-	// Parent is the parent device of this blockdevice, if it exists.
-	// It will always be a single device.
-	Parent string
-
-	// Partitions is the list of partitions(again blockdevices) for
-	// this blockdevice, if it exists.
-	Partitions []string
-
-	// Holders is the list of blockdevices that are held by this blockdevice.
-	// eg: sda1 can hold dm-0. Then the list of sda1 will contain dm-0.
-	Holders []string
-
-	// Slaves is the list of blockdevices to which this blockdevice is a slave.
-	// Slaves are represented in SysFS under the block device layer,
-	// so they are dependent on the device
-	// eg: dm-0 is a slave to sda1. Then the list of dm-0 will contain sda1
-	Slaves []string
+	// DependentDevices stores the dependent devices
+	DependentDevices DependentBlockDevices
 
 	// TemperatureInfo stores the temperature information of the drive
 	TemperatureInfo TemperatureInformation
@@ -296,6 +281,28 @@ type PartitionInformation struct {
 	PartitionTableType string
 }
 
+// DependentBlockDevices contains path of all devices that are
+// related to this BlockDevice
+type DependentBlockDevices struct {
+	// Parent is the parent device of this blockdevice, if it exists.
+	// It will always be a single device.
+	Parent string
+
+	// Partitions is the list of partitions(again blockdevices) for
+	// this blockdevice, if it exists.
+	Partitions []string
+
+	// Holders is the list of blockdevices that are held by this blockdevice.
+	// eg: sda1 can hold dm-0. Then the list of sda1 will contain dm-0.
+	Holders []string
+
+	// Slaves is the list of blockdevices to which this blockdevice is a slave.
+	// Slaves are represented in SysFS under the block device layer,
+	// so they are dependent on the device
+	// eg: dm-0 is a slave to sda1. Then the list of dm-0 will contain sda1
+	Slaves []string
+}
+
 // Status is used to represent the status of the blockdevice
 type Status struct {
 	// State is the state of this BD like Active(Online), Inactive(Offline) or
@@ -323,3 +330,6 @@ const (
 	// claiming
 	Unclaimed string = "Unclaimed"
 )
+
+// Hierarchy is the block device hierarchy on the system
+type Hierarchy map[string]BlockDevice
--- a/blockdevice/blockdevice.go
+++ b/blockdevice/blockdevice.go
@@ -331,5 +331,11 @@ const (
 	Unclaimed string = "Unclaimed"
 )
 
-// Hierarchy is the block device hierarchy on the system
+// Hierarchy is the block device hierarchy on the system.
+// This map will be used as in memory cache for storing the current state of
+// all block devices on the system. This cache helps to get the details of any
+// dependent blockdevices from a device.
+// eg: If at a certain point, we are processing sda1 and we need to get the UUID
+// of its parent device. We will get /dev/sda is the parent of /dev/sda1.
+// This will be used to query the cache and get/generate the UUID of /dev/sda.
 type Hierarchy map[string]BlockDevice
--- a/cmd/ndm_daemonset/controller/blockdevicestore.go
+++ b/cmd/ndm_daemonset/controller/blockdevicestore.go
@@ -97,9 +97,8 @@ func (c *Controller) UpdateBlockDevice(blockDevice apis.BlockDevice, oldBlockDev
 		}
 	}
 
-	blockDeviceCopy.ObjectMeta = mergeMetadata(blockDeviceCopy.ObjectMeta, oldBlockDevice.ObjectMeta)
-	blockDeviceCopy.Spec.ClaimRef = oldBlockDevice.Spec.ClaimRef
-	blockDeviceCopy.Status.ClaimState = oldBlockDevice.Status.ClaimState
+	blockDeviceCopy = mergeBlockDeviceData(*blockDeviceCopy, *oldBlockDevice)
+
 	err = c.Clientset.Update(context.TODO(), blockDeviceCopy)
 	if err != nil {
 		klog.Errorf("eventcode=%s msg=%s : %v rname=%v",
@@ -258,6 +257,29 @@ func (c *Controller) MarkBlockDeviceStatusToUnknown() {
 				blockDeviceCopy.ObjectMeta.Name)
 		}
 	}
+}
+
+// mergeBlockDeviceData merges the data from BlockDevice resource available in etcd
+// with the system generated BlockDevice information
+// If the device is in use, then only the capacity, node attributes, path, devlinks
+// and state will be updated. This is because, these are the fields relevant even if
+// the device is in use.
+func mergeBlockDeviceData(newBD, oldBD apis.BlockDevice) *apis.BlockDevice {
+	oldBD.TypeMeta = newBD.TypeMeta
+	oldBD.ObjectMeta = mergeMetadata(newBD.ObjectMeta, oldBD.ObjectMeta)
+	// if the device is in use, only the below fields will be updated.
+	if oldBD.Status.ClaimState != apis.BlockDeviceUnclaimed {
+		klog.V(4).Infof("device: %s is in use, updating only relevant fields", newBD.Spec.Path)
+		oldBD.Spec.NodeAttributes = newBD.Spec.NodeAttributes
+		oldBD.Spec.Capacity.Storage = newBD.Spec.Capacity.Storage
+		oldBD.Spec.Path = newBD.Spec.Path
+		oldBD.Spec.DevLinks = newBD.Spec.DevLinks
+		oldBD.Status.State = newBD.Status.State
+	} else {
+		oldBD.Spec = newBD.Spec
+		oldBD.Status = newBD.Status
+	}
+	return &oldBD
 }
 
 // mergeMetadata merges oldMetadata with newMetadata. It takes old metadata and
--- a/cmd/ndm_daemonset/controller/blockdevicestore.go
+++ b/cmd/ndm_daemonset/controller/blockdevicestore.go
@@ -164,9 +164,9 @@ func (c *Controller) DeleteBlockDevice(name string) {
 		"ndm.blockdevice.delete.success", "Deleted blockdevice object ", name)
 }
 
-// ListBlockDeviceResource queries the etcd for the devices for the host/node
+// ListBlockDeviceResource queries the etcd for the devices
 // and returns list of blockdevice resources.
-func (c *Controller) ListBlockDeviceResource() (*apis.BlockDeviceList, error) {
+func (c *Controller) ListBlockDeviceResource(listAll bool) (*apis.BlockDeviceList, error) {
 
 	blockDeviceList := &apis.BlockDeviceList{
 		TypeMeta: metav1.TypeMeta{
@@ -174,8 +174,11 @@ func (c *Controller) ListBlockDeviceResource() (*apis.BlockDeviceList, error) {
 			APIVersion: "openebs.io/v1alpha1",
 		},
 	}
-	filter := KubernetesHostNameLabel + "=" + c.NodeAttributes[HostNameKey]
-	filter = filter + "," + NDMManagedKey + "!=" + FalseString
+	filter := NDMManagedKey + "!=" + FalseString
+	// whether to list all devices in the cluster, or the devices that belong to this node
+	if !listAll {
+		filter = filter + "," + KubernetesHostNameLabel + "=" + c.NodeAttributes[HostNameKey]
+	}
 	opts := &client.ListOptions{}
 	_ = opts.SetLabelSelector(filter)
 	err := c.Clientset.List(context.TODO(), opts, blockDeviceList)
@@ -213,7 +216,7 @@ func (c *Controller) GetExistingBlockDeviceResource(blockDeviceList *apis.BlockD
 // system that will be marked as inactive.
 func (c *Controller) DeactivateStaleBlockDeviceResource(devices []string) {
 	listDevices := append(devices, GetActiveSparseBlockDevicesUUID(c.NodeAttributes[HostNameKey])...)
-	blockDeviceList, err := c.ListBlockDeviceResource()
+	blockDeviceList, err := c.ListBlockDeviceResource(false)
 	if err != nil {
 		klog.Error(err)
 		return
@@ -241,7 +244,7 @@ func (c *Controller) PushBlockDeviceResource(oldBlockDevice *apis.BlockDevice,
 // MarkBlockDeviceStatusToUnknown makes state of all resources owned by node unknown
 // This will call as a cleanup process before shutting down.
 func (c *Controller) MarkBlockDeviceStatusToUnknown() {
-	blockDeviceList, err := c.ListBlockDeviceResource()
+	blockDeviceList, err := c.ListBlockDeviceResource(false)
 	if err != nil {
 		klog.Error(err)
 		return
--- a/cmd/ndm_daemonset/controller/blockdevicestore.go
+++ b/cmd/ndm_daemonset/controller/blockdevicestore.go
@@ -165,6 +165,8 @@ func (c *Controller) DeleteBlockDevice(name string) {
 
 // ListBlockDeviceResource queries the etcd for the devices
 // and returns list of blockdevice resources.
+// if listAll = true, all the BlockDevices in the cluster will be listed,
+// else only devices present in this node will be listed.
 func (c *Controller) ListBlockDeviceResource(listAll bool) (*apis.BlockDeviceList, error) {
 
 	blockDeviceList := &apis.BlockDeviceList{
--- a/cmd/ndm_daemonset/controller/probe.go
+++ b/cmd/ndm_daemonset/controller/probe.go
@@ -101,7 +101,6 @@ func (c *Controller) ListProbe() []*Probe {
 // FillBlockDeviceDetails lists registered probes and fills details from each probe
 func (c *Controller) FillBlockDeviceDetails(blockDevice *blockdevice.BlockDevice) {
 	blockDevice.NodeAttributes = c.NodeAttributes
-	blockDevice.DeviceAttributes.DeviceType = NDMDefaultDiskType
 	probes := c.ListProbe()
 	for _, probe := range probes {
 		probe.FillBlockDeviceDetails(blockDevice)
--- a/cmd/ndm_daemonset/probe/eventhandler.go
+++ b/cmd/ndm_daemonset/probe/eventhandler.go
@@ -17,8 +17,12 @@ limitations under the License.
 package probe
 
 import (
+	"github.com/openebs/node-disk-manager/blockdevice"
 	"github.com/openebs/node-disk-manager/cmd/ndm_daemonset/controller"
+	"github.com/openebs/node-disk-manager/pkg/features"
+	"github.com/openebs/node-disk-manager/pkg/partition"
 	libudevwrapper "github.com/openebs/node-disk-manager/pkg/udev"
+	"github.com/openebs/node-disk-manager/pkg/util"
 	"k8s.io/klog"
 )
 
@@ -48,13 +52,53 @@ func (pe *ProbeEvent) addBlockDeviceEvent(msg controller.EventMessage) {
 	isErrorDuringUpdate := false
 	// iterate through each block device and perform the add/update operation
 	for _, device := range msg.Devices {
-		klog.Infof("Processing details for %s", device.UUID)
+		klog.Infof("Processing details for %s", device.DevPath)
 		pe.Controller.FillBlockDeviceDetails(device)
 		// if ApplyFilter returns true then we process the event further
 		if !pe.Controller.ApplyFilter(device) {
 			continue
 		}
-		klog.Infof("Processed details for %s : %s", device.DevPath, device.UUID)
+		klog.Infof("Processed details for %s", device.DevPath)
+
+		// if GPTBasedUUID need to be used, generate the UUID,
+		// if UUID cannot be generated create a GPT partition
+		if pe.Controller.FeatureGates.IsEnabled(features.GPTBasedUUID) {
+			if len(device.Partitions) > 0 {
+				klog.Info("device has partitions. not creating blockdevice resource")
+				continue
+			}
+			if len(device.Holders) > 0 {
+				klog.Info("device has holder devices, not creating blockdevice resource")
+				continue
+			}
+			uuid, ok := generateUUID(*device)
+			// manaully create a single partition on the device
+			if !ok {
+				klog.Info("starting to create partitions")
+				d := partition.Disk{
+					DevPath:          device.DevPath,
+					DiskSize:         device.Capacity.Storage,
+					LogicalBlockSize: uint64(device.DeviceAttributes.LogicalBlockSize),
+				}
+				if err := d.CreatePartitionTable(); err != nil {
+					klog.Errorf("error creating partition table for %s, %v", device.DevPath, err)
+					continue
+				}
+				if err = d.AddPartition(); err != nil {
+					klog.Errorf("error creating partition for %s, %v", device.DevPath, err)
+					continue
+				}
+				if err = d.ApplyPartitionTable(); err != nil {
+					klog.Errorf("error writing partition data to %s, %v", device.DevPath, err)
+					continue
+				}
+				klog.Infof("created new partition in %s", device.DevPath)
+				continue
+			}
+			klog.Infof("generated UUID: %s for device: %s", uuid, device.DevPath)
+			device.UUID = uuid
+		}
+
 		deviceInfo := pe.Controller.NewDeviceInfoFromBlockDevice(device)
 
 		existingBlockDeviceResource := pe.Controller.GetExistingBlockDeviceResource(bdList, deviceInfo.UUID)
@@ -72,32 +116,40 @@ func (pe *ProbeEvent) addBlockDeviceEvent(msg controller.EventMessage) {
 
 // deleteBlockDeviceEvent deactivate blockdevice resource using uuid from etcd
 func (pe *ProbeEvent) deleteBlockDeviceEvent(msg controller.EventMessage) {
-	blockDeviceOk := pe.deleteBlockDevice(msg)
-
-	// when one disk is removed from node and entry related to
-	// that disk is not present in etcd,  in that case it
-	// again rescan full system and update etcd accordingly.
-	if !blockDeviceOk {
-		go pe.initOrErrorEvent()
-	}
-}
-
-func (pe *ProbeEvent) deleteBlockDevice(msg controller.EventMessage) bool {
 	bdList, err := pe.Controller.ListBlockDeviceResource()
 	if err != nil {
 		klog.Error(err)
-		return false
 	}
-	ok := true
-	for _, diskDetails := range msg.Devices {
-		existingBlockDeviceResource := pe.Controller.GetExistingBlockDeviceResource(bdList, diskDetails.UUID)
+
+	isDeactivated := true
+	isGPTBasedUUIDEnabled := pe.Controller.FeatureGates.IsEnabled(features.GPTBasedUUID)
+
+	for _, device := range msg.Devices {
+		// create the new UUID for removing the device
+		if isGPTBasedUUIDEnabled {
+			uuid, ok := generateUUID(*device)
+			if !ok {
+				klog.Info("could not recreate UUID while removing device")
+				continue
+			}
+			// use the new UUID for deactivating the blockdevice
+			device.UUID = uuid
+		}
+
+		existingBlockDeviceResource := pe.Controller.GetExistingBlockDeviceResource(bdList, device.UUID)
 		if existingBlockDeviceResource == nil {
-			ok = false
+			isDeactivated = false
 			continue
 		}
 		pe.Controller.DeactivateBlockDevice(*existingBlockDeviceResource)
 	}
-	return ok
+
+	// TODO @akhilerm check if rescan is required every time and take appropriate actions
+	//
+	// rescan only if GPT based UUID is disabled.
+	if !isDeactivated && !isGPTBasedUUIDEnabled {
+		go pe.initOrErrorEvent()
+	}
 }
 
 // initOrErrorEvent rescan system and update disk resource this is
@@ -109,4 +161,38 @@ func (pe *ProbeEvent) initOrErrorEvent() {
 	if err != nil {
 		klog.Error(err)
 	}
+}
+
+// generateUUID creates a new UUID based on the algorithm proposed in
+// https://github.com/openebs/openebs/pull/2666
+func generateUUID(bd blockdevice.BlockDevice) (string, bool) {
+	var ok bool
+	var uuidField, uuid string
+
+	// select the field which is to be used for generating UUID
+	switch {
+	case bd.DeviceAttributes.DeviceType == libudevwrapper.UDEV_PARTITION:
+		klog.Infof("device(%s) is a partition, using partition UUID: %s", bd.DevPath, bd.PartitionInfo.PartitionEntryUUID)
+		uuidField = bd.PartitionInfo.PartitionEntryUUID
+		ok = true
+	case len(bd.DeviceAttributes.WWN) > 0:
+		klog.Infof("device(%s) has a WWN, using WWN: %s", bd.DevPath, bd.DeviceAttributes.WWN)
+		uuidField = bd.DeviceAttributes.WWN
+		ok = true
+	case len(bd.PartitionInfo.PartitionTableUUID) > 0:
+		klog.Infof("device(%s) has a partition table, using partition table UUID: %s", bd.DevPath, bd.PartitionInfo.PartitionTableUUID)
+		uuidField = bd.PartitionInfo.PartitionTableUUID
+		ok = true
+	case len(bd.FSInfo.FileSystemUUID) > 0:
+		klog.Infof("device(%s) has a filesystem, using filesystem UUID: %s", bd.DevPath, bd.FSInfo.FileSystemUUID)
+		uuidField = bd.FSInfo.FileSystemUUID
+		ok = true
+	}
+
+	if ok {
+		uuid = libudevwrapper.NDMBlockDevicePrefix + util.Hash(uuidField)
+		klog.Infof("generated uuid: %s for device: %s", uuid, bd.DevPath)
+	}
+
+	return uuid, ok
 }
--- a/cmd/ndm_daemonset/probe/eventhandler.go
+++ b/cmd/ndm_daemonset/probe/eventhandler.go
@@ -72,9 +72,9 @@ func (pe *ProbeEvent) addBlockDeviceEvent(msg controller.EventMessage) {
 				continue
 			}
 			uuid, ok := generateUUID(*device)
-			// manaully create a single partition on the device
+			// manually create a single partition on the device
 			if !ok {
-				klog.Info("starting to create partitions")
+				klog.Info("starting to create partition")
 				d := partition.Disk{
 					DevPath:          device.DevPath,
 					DiskSize:         device.Capacity.Storage,
--- a/cmd/ndm_daemonset/probe/eventhandler.go
+++ b/cmd/ndm_daemonset/probe/eventhandler.go
@@ -49,6 +49,9 @@ func (pe *ProbeEvent) addBlockDeviceEvent(msg controller.EventMessage) {
 		go pe.initOrErrorEvent()
 		return
 	}
+
+	isGPTBasedUUIDEnabled := pe.Controller.FeatureGates.IsEnabled(features.GPTBasedUUID)
+
 	isErrorDuringUpdate := false
 	// iterate through each block device and perform the add/update operation
 	for _, device := range msg.Devices {
@@ -62,7 +65,7 @@ func (pe *ProbeEvent) addBlockDeviceEvent(msg controller.EventMessage) {
 
 		// if GPTBasedUUID need to be used, generate the UUID,
 		// if UUID cannot be generated create a GPT partition
-		if pe.Controller.FeatureGates.IsEnabled(features.GPTBasedUUID) {
+		if isGPTBasedUUIDEnabled {
 			if len(device.Partitions) > 0 {
 				klog.Info("device has partitions. not creating blockdevice resource")
 				continue
@@ -97,6 +100,14 @@ func (pe *ProbeEvent) addBlockDeviceEvent(msg controller.EventMessage) {
 			}
 			klog.Infof("generated UUID: %s for device: %s", uuid, device.DevPath)
 			device.UUID = uuid
+		}
+
+		// if GPTBasedUUID is disabled and the device type is partition,
+		// the event can be skipped.
+		if !isGPTBasedUUIDEnabled &&
+			device.DeviceAttributes.DeviceType == libudevwrapper.UDEV_PARTITION {
+			klog.Info("GPTBasedUUID disabled. skip creating block device resource for partition.")
+			continue
 		}
 
 		deviceInfo := pe.Controller.NewDeviceInfoFromBlockDevice(device)
--- a/cmd/ndm_daemonset/probe/eventhandler.go
+++ b/cmd/ndm_daemonset/probe/eventhandler.go
@@ -168,8 +168,17 @@ func generateUUID(bd blockdevice.BlockDevice) (string, bool) {
 	var uuidField, uuid string
 
 	// select the field which is to be used for generating UUID
+	//
+	// Serial number is not used for UUID generation. This is because serial number is not
+	// unique in some cloud environments. For example, in GCP the serial number is
+	// configurable by the --device-name flag while attaching the disk.
+	// If this flag is not provided, GCP automatically assigns the serial number
+	// which is unique only to the node.
 	switch {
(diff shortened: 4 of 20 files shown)


akhilerm:
Final implementation in #378

cmd/ndm_daemonset/probe/udevprobe.go
```diff
@@ -172,7 +181,13 @@ func (up *udevProbe) scan() error {
...
 		newUdevice.UdevDeviceUnref()
 	}
-	up.controller.DeactivateStaleBlockDeviceResource(disksUid)
+	if !up.controller.FeatureGates.IsEnabled(features.GPTBasedUUID) {
+		up.controller.DeactivateStaleBlockDeviceResource(disksUid)
+	} else {
+		// TODO deactivate all the stale blockdevices
+		// can deactivate all the blockdevices and they will be marked as active once
+		// the event is processed.
+	}
```
akhilerm: > hmm, deactivating all of them seems to be ok as if it was graceful shutdown they will be Inactive anyway. I don't see any corner case just that it will be extra etcd call for the devices that are currently attached one for making them Incative and other for making them active. What is the problem in calling DeactivateStaleBlockDeviceResource here? We can have disksUid map created for feature GPTBasedUUID) and pass it here?

@pawanpraka1  quoting your comment from #378 . 

We are generating the UUID at the last step, just before updating the resource into etcd. So, at this point after detecting a device, we don't know its UUID.

The reason for generating UUID in `eventhandler` and not here.
The `syspath` & `devpath` will serve the purpose of unique identification within the scope of the host. Only when we -need to jump into cluster wide scope, we need the actual UUID. Therefore, the UUID is created just before updating to etcd. This will also prevent us from pushing too much changes of creating partitions also, into the device detection code. 

Since UUID is not available at this point, I thought may be we could deactivate all resources belonging to this host, and then let the udev update with the details accordingly.

cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -63,32 +107,40 @@ func (pe *ProbeEvent) addBlockDeviceEvent(msg controller.EventMessage) {
...
 			continue
 		}
 		pe.Controller.DeactivateBlockDevice(*existingBlockDeviceResource)
 	}
-	return ok
+
+	// TODO @akhilerm check if rescan is required every time and take appropriate actions
+	//
+	// rescan only if GPT based UUID is disabled.
+	if !isDeactivated && !isGPTBasedUUIDEnabled {
```
akhilerm: @pawanpraka1  I also have a concern or rescanning everytime, when a device which was not used is removed. Even in case of filtered out devices. This can happen. IMO we should avoid these also. 
Need your thoughts on why it was introduced in the first place.

pawanpraka1: yes, the idea was if any thing fails, scan the entire disks. For filtered disks, this was probably a bug as we look for that device in etcd, which will not be there so we scan it again unnecessary.

cmd/ndm_daemonset/probe/udevprobe.go
pawanpraka1: hmmmm, I am fine with deactivating everything before processing anything.

akhilerm: ok. then will update with the changes.

cmd/ndm_daemonset/probe/eventhandler.go
akhilerm: ok. then will update the code so that a rescan is not performed, if there is any error during `deleteBlockDeviceEvent`

cmd/ndm_daemonset/probe/udevprobe.go
akhilerm: done.

cmd/ndm_daemonset/probe/eventhandler.go
akhilerm: done.

cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -47,13 +51,53 @@ func (pe *ProbeEvent) addBlockDeviceEvent(msg controller.EventMessage) {
...
+				klog.Info("device has partitions. not creating blockdevice resource")
+				continue
+			}
+			if len(device.Holders) > 0 {
+				klog.Info("device has holder devices, not creating blockdevice resource")
+				continue
+			}
+			uuid, ok := generateUUID(*device)
+			// manaully create a single partition on the device
+			if !ok {
```
gila: There is no need to create a partition -- creating a GPT label should be fine. The purpose is to ID the disk if we don't have HW identifiers and there are no existing GPT labels or FS ID's, right?

cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -47,13 +51,53 @@ func (pe *ProbeEvent) addBlockDeviceEvent(msg controller.EventMessage) {
...
 			continue
 		}
-		klog.Infof("Processed details for %s : %s", device.DevPath, device.UUID)
+		klog.Infof("Processed details for %s", device.DevPath)
+
+		// if GPTBasedUUID need to be used, generate the UUID,
+		// if UUID cannot be generated create a GPT partition
+		if pe.Controller.FeatureGates.IsEnabled(features.GPTBasedUUID) {
+			if len(device.Partitions) > 0 {
+				klog.Info("device has partitions. not creating blockdevice resource")
```
gila: Should this not be, not creating GPT labels? I'm skeptical as to why this is logged in the fist place. is klog.debug() not more appropriate?

cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -63,32 +107,38 @@ func (pe *ProbeEvent) addBlockDeviceEvent(msg controller.EventMessage) {
...
-func (pe *ProbeEvent) deleteBlockDevice(msg controller.EventMessage) bool {
 	bdList, err := pe.Controller.ListBlockDeviceResource()
 	if err != nil {
 		klog.Error(err)
-		return false
 	}
-	ok := true
-	for _, diskDetails := range msg.Devices {
-		existingBlockDeviceResource := pe.Controller.GetExistingBlockDeviceResource(bdList, diskDetails.UUID)
+
```
gila: I'm not sure I understand what is happening here. It looks like, when we device gets deleted, it gets a new label?

cmd/ndm_daemonset/probe/udevprobe.go
```diff
@@ -142,20 +144,27 @@ func (up *udevProbe) scan() error {
...
 		}
 		if newUdevice.IsDisk() || newUdevice.IsParitition() {
-			uuid := newUdevice.GetUid()
-			disksUid = append(disksUid, uuid)
 			deviceDetails := &blockdevice.BlockDevice{}
-			deviceDetails.UUID = uuid
+			if up.controller.FeatureGates.IsEnabled(features.GPTBasedUUID) {
+				// all fields that is used to generate UUID to be filled whenever the event is
+				// generated. This is required so that even if none of the probes work, the
+				// system should be able to generate the UUID of the blockdevice
```
gila: Comment is not clear to me sorry.

pkg/partition/partition.go
```diff
@@ -0,0 +1,109 @@
...
+		LogicalSectorSize: int(d.LogicalBlockSize),
+		ProtectiveMBR:     true,
+	}
+	return nil
+}
+
+// AddPartition is used to add a partition to the partition table.
+// Currently only a single partition can be created. i.e The method can be called only once for a disk
+// TODO: @akhilerm, add method to create partiton with given size
+func (d *Disk) AddPartition() error {
```
gila: As mentioned in an other comment and considering the restriction you state within this comment: why not simply label it and be done with it. IOW do not create any partitions at all?

cmd/ndm_daemonset/probe/eventhandler.go
akhilerm: Yes, the purpose is to ID the disk which do not have hardware identifiers. But, if we just write the GPT label to a disk suppose `/dev/sda`, then the consumer of that device can still overwrite the GPT label that we have written and we loose the unique identifier.

If we write the GPT label and then create a single partition that spans the entire drive, the consumers will be made available only the partition `/dev/sda1`, and our  GPT label will remain safe from overwrites.

cmd/ndm_daemonset/probe/eventhandler.go
akhilerm: I will update the log level to `klog.V(4).Info`, as klog explicity support a debug level.

cmd/ndm_daemonset/probe/eventhandler.go
akhilerm: Nope. When a udev `remove` event occurs for a device we need to mark that BlockDevice as Inactive. So we will try to generate the UUID for the device which was removed. The label won't be recreated, we use the existing label and use it to identify the device which was removed.

cmd/ndm_daemonset/probe/udevprobe.go
akhilerm: `WWN`, `PartitionTableUUID/GPTLabel`, `PartitionUUID`, `FileSystemUUID` and `DeviceType` are the fields we use to generate the UUID. These fields will be fetched from the udev event itself. This is to guarantee that we donot need to rely on any other probes to fill in those details which are ciritcal for device identification.

cmd/ndm_daemonset/probe/eventhandler.go
gila: That can always happen and you have no way of avoiding that. If you "fdisk" a disk with no partitions, and it has a GPT label, the system will write the new partition table into the existing label -- it will not wipe out the existing GPT label.

cmd/ndm_daemonset/probe/udevprobe.go
gila: May I suggest you replace what you just said with the code comment?

cmd/ndm_daemonset/probe/eventhandler.go
akhilerm: But it is not guaranteed that the application will create a partition right. I will list out an example scenario, where I thought partition is required.
1. NDM creates a GPT label on to the block device(`/dev/sdb`) for identification
2. The blockdevice is claimed by an application.
3. The application formats the disk with a filesystem, say ext4. Then the GPT label will be removed right.
This means we no longer can identify the disk.

gila: You will have the FSID.

cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -101,3 +151,37 @@ func (pe *ProbeEvent) initOrErrorEvent() {
...
+		uuidField = bd.PartitionInfo.PartitionTableUUID
+		ok = true
+	case len(bd.FSInfo.FileSystemUUID) > 0:
+		klog.Infof("device(%s) has a filesystem, using filesystem UUID: %s", bd.DevPath, bd.FSInfo.FileSystemUUID)
+		uuidField = bd.FSInfo.FileSystemUUID
+		ok = true
+	}
+
+	if ok {
+		uuid = libudevwrapper.NDMBlockDevicePrefix + util.Hash(uuidField)
```
gila: Can you remove the hash? I don't think it is needed with the current approach.

cmd/ndm_daemonset/probe/eventhandler.go
gila: Ok, I see. If you remove the hash generation you can just delete the device directly.

cmd/ndm_daemonset/probe/eventhandler.go
akhilerm: But that means, the unique ID we have used has changed(GPT Label -> FSUID). What we need is something that persists irrespective of what the user does.

cmd/ndm_daemonset/probe/eventhandler.go
akhilerm: One purpose of using hash was to get a common format for the UUID, irrespective of the field we use to generate the UUID.
eg: WWN can be of the format : `0x500a0751e2105248`
  FS_UUID : `d716ed29-1f2c-461a-aac7-fd7bbb835b33`
PartitionLabel : `cc5e33c5-adba-cf44-a79d-ac3c37b942a2`

using a hash will give a consistent name like `blockdevice-md5(WWN/FS_UUID/GPTLabel)`
Are there any issues with using hash @gila ?

gila: When I have the WWN or the UUID I can relate to that physically on the host and any other tools/SW I'm using. Those hashes are not used outside of NDM, and if I have an error on the host (showing up in the logs let say) I will see things like UUIDs and FS-IDS and not the hashes that are generated. 

Also, if given just a hash, I can not determine what the input would have been. IOW I can't go back to the WWN which makes it very unpractical IMHO.

BTW -- if a disk has a WWN, I'm pretty sure it also has a UUID, can you verify this? If that is true, then we can always show the UUID and the problem goes away?

cmd/ndm_daemonset/probe/eventhandler.go
gila: Yes, we need something that is not tied directly to what the user does. But as we know for a very *small* set of cases this is not the truth, unfortunately. 

For cases like these, I think we should follow what the user does, and not assume that the user will never change the FS type, partition layout or anything like that. How will you even prevent the user from not reformatting or repartitioning the disk?

cmd/ndm_daemonset/probe/eventhandler.go
akhilerm: @gila the fields that we use to generate the UUID (i.e the value which was hashed) will be stored in `.spec.details` of the blockdevice object. So you can always associate it. The sole purpose of using the hash was to expose consistently formatted name to the users.

> BTW -- if a disk has a WWN, I'm pretty sure it also has a UUID, can you verify this?
What do you mean by the UUID here?

vishnuitta:
Removed ReviewRequest on me as I'm NOT actively looking at this and so I don't want to block PR..

cmd/ndm_daemonset/probe/eventhandler.go
kmova: Had a chat with @gila and @akhilerm offline on this comment. 

As of today, there is no better option available than to create a partition to identify the virtual disk uniquely across reboots. In case the user after taking the block device completely reformats the disk, there is no way for the NDM to detect and update its resources. This will be added as a known limitation with the current version of NDM.

cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -110,3 +160,37 @@ func (pe *ProbeEvent) initOrErrorEvent() {
...
+		ok = true
+	case len(bd.DeviceAttributes.WWN) > 0:
+		klog.Infof("device(%s) has a WWN, using WWN: %s", bd.DevPath, bd.DeviceAttributes.WWN)
+		uuidField = bd.DeviceAttributes.WWN
+		ok = true
+	case len(bd.PartitionInfo.PartitionTableUUID) > 0:
+		klog.Infof("device(%s) has a partition table, using partition table UUID: %s", bd.DevPath, bd.PartitionInfo.PartitionTableUUID)
+		uuidField = bd.PartitionInfo.PartitionTableUUID
+		ok = true
+	case len(bd.FSInfo.FileSystemUUID) > 0:
```
kmova: Should there be a case of adding serial number also to this list or is serial number not a unique field?

cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -110,3 +160,37 @@ func (pe *ProbeEvent) initOrErrorEvent() {
...
+	switch {
+	case bd.DeviceAttributes.DeviceType == libudevwrapper.UDEV_PARTITION:
+		klog.Infof("device(%s) is a partition, using partition UUID: %s", bd.DevPath, bd.PartitionInfo.PartitionEntryUUID)
+		uuidField = bd.PartitionInfo.PartitionEntryUUID
+		ok = true
+	case len(bd.DeviceAttributes.WWN) > 0:
+		klog.Infof("device(%s) has a WWN, using WWN: %s", bd.DevPath, bd.DeviceAttributes.WWN)
+		uuidField = bd.DeviceAttributes.WWN
+		ok = true
+	case len(bd.PartitionInfo.PartitionTableUUID) > 0:
```
kmova: Can you add a comment on how this case is different from the first case at line 172?

cmd/ndm_daemonset/probe/eventhandler.go
akhilerm: Serial number is not unique in some cloud environments. For example, in GCP the serial number is configurable by the `--device-name` flag while attaching the disk. If this flag is not provided, GCP automatically assigns the serial number which is unique only to the node.Hence we cannot use serial number.

kmova: Probably include the above information also as a comment. It will be helpful for future refactoring.

akhilerm: Done.

cmd/ndm_daemonset/probe/eventhandler.go
akhilerm: Comments added.

cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -178,6 +187,10 @@ func generateUUID(bd blockdevice.BlockDevice) (string, bool) {
 		uuidField = bd.DeviceAttributes.WWN
 		ok = true
 	case len(bd.PartitionInfo.PartitionTableUUID) > 0:
+		// PartitionTable UUID is used if the disk contains only the partition table and not a partition. This is a very rare
```
kmova: If the NDM identifies the disk using partition table uuid:
(a) Will the device path still be /dev/sda?
(b) If given to Local PV. Local PV will format the device with filesystem. On reboot / restart of the ndm will uuid be generated correctly?
(c) If given to cStor, cStor will create partitions. On reboot/restart of NDM, will uuid be detected correctly?

akhilerm: yes, a) the device path will still be `/dev/sda` 
b) If formatted it won't work,
c) If partitions are created, exisiting partition table UUID will be used.

In that case, shall we update the design to remove partition table UUID based identification. and always use partition entry

akhilerm: update code to remove `PartitionTableUUID` from algorithm used for UUID generation.

cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -186,14 +195,6 @@ func generateUUID(bd blockdevice.BlockDevice) (string, bool) {
...
-		// PartitionTable UUID is used if the disk contains only the partition table and not a partition. This is a very rare
-		// case where the user creates the partition table, but does not create any partition on the disk. In such a case
-		// partition table UUID will be used. In all other cases where partition table UUID is present, the disk will also
-		// have partition entries and it will be used.
-		klog.Infof("device(%s) has a partition table, using partition table UUID: %s", bd.DevPath, bd.PartitionInfo.PartitionTableUUID)
-		uuidField = bd.PartitionInfo.PartitionTableUUID
-		ok = true
 	case len(bd.FSInfo.FileSystemUUID) > 0:
 		klog.Infof("device(%s) has a filesystem, using filesystem UUID: %s", bd.DevPath, bd.FSInfo.FileSystemUUID)
 		uuidField = bd.FSInfo.FileSystemUUID
```
kmova: One more query @akhilerm, consider this case:

- User creates a GKE cluster with Local SSDs, that can formated and mounted by default. 
- User installs openebs
- NDM discovers the block devices using the FS ID
- Users wants to install cStor on those block devices, so user follows the troubleshooting guide and realizes that blockdevices should not have a file system. So "wipefs -a` is executed. 

What happens to those block devices? Do we end up creating stale block devices? Not as part of this PR, but should we considering a Block Device, that is unknown state but is not claimed? 

In the above scenario - it is possible that BDC was already attached to the Old BDs, but then due to user erasing the Filesystem, those BDs will go to unknown state.  Or I think cStor doesn't claim devices that already have filesystems on it, which is better.

akhilerm: - cStor will pick up only devices which does not have a filesystem to create pools.

But the case that you mentioned can happen, to prevent it while creating BlockDeviceClaim, the user can specify the `VolumeMode` for the BlockDevice to claim. If VolumeMode is specified as `Block`, only devices without any filesystem will be claimed. 
Currently by default, if volume mode is not specified, any block device can be claimed. I think, if we change the default behaviour to claim only blockdevices in `Block` mode, this issue will not happen.

pkg/partition/partition.go
```diff
@@ -0,0 +1,109 @@
...
+	var startSector, endSector uint64
+	if len(d.table.Partitions) == 0 {
+		// First sector of partition is aligned at 1MiB
+		startSector = (GPTPartitionStartByte) / d.LogicalBlockSize
+	}
+	// 2 blocks for LBA 0 and LBA 1
+	gptHeaderInSectors := BytesRequiredForGPTHeader/d.LogicalBlockSize + 2
```
vishnuitta: is LBA 0 and LBA 1 are at the end of the disk?
endSector doesn't seem to be correct

akhilerm: > is LBA 0 and LBA 1 are at the end of the disk?

No.

We are calculating the size of the GPT Header here. Assume the size of GPT Header which is found at the starting of disk is `X`. In GPT schema, the partition header and partition entries are backed up on the last LBAs of the disk also. So for storing this backup data we need `X` size at the end of the disk. 
Therefore, the end sector of the partition = last sector on disk - `X`

More details available [here](https://metebalci.com/blog/a-quick-tour-of-guid-partition-table-gpt/)

vishnuitta: it seems GPT Partition Table (LBA 1) and GPT Partition Entry Array (BytesRequiredForGPTHeader/LogicalBlockSize) are stored at the end also..
so, end sector should be (last disk sector - (BytesRequiredForGPTHeader/LogicalBlockSize) - 1)
It seems - we are missing one block here..
endSector can be (d.DiskSize/d.LogicalBlockSize) - gptHeaderInSectors + 1 (This extra 1 is for LBA 0 which is added in gptHeaderInSectors.

vishnuitta: thats a very good link @akhilerm .. however, this one block should NOT matter.. its ok to lose a block as long as data is fine.. :)

pkg/partition/partition.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// AddPartition is used to add a partition to the partition table.
+// Currently only a single partition can be created. i.e The method can be called only once for a disk
+// TODO: @akhilerm, add method to create partiton with given size
+func (d *Disk) AddPartition() error {
+	var startSector, endSector uint64
+	if len(d.table.Partitions) == 0 {
+		// First sector of partition is aligned at 1MiB
+		startSector = (GPTPartitionStartByte) / d.LogicalBlockSize
```
vishnuitta, line 93: almost all the time GPTPartitionStartByte is divisible by LogicalBlockSize, however how about making it as:
if (GPTPartitionStartByte % d.LogicalBlockSize == 0)
startSector = GPTPartitionStartByte / d.LogicalBlockSize
else
startSector = GPTPartitionStartByte / d.LogicalBlockSize + 1

akhilerm: @vishnuitta , the startbyte of partitions should be always aligned to logical blocks. Otherwise it will impact performance. (quoted from fdisk). This is the reason, the partition start is always set at 1MB, so that block sizes of 512, 4k, 16k or something that comes up later can also be handled without degrading performance.

pkg/partition/partition.go
akhilerm: > so, end sector should be (last disk sector - (BytesRequiredForGPTHeader/LogicalBlockSize) - 1)

this is correct. But the LBA ranges from 0 to (last block - 1) right? Therefore when we calculate, its correct, we won't miss any blocks.

pkg/partition/partition.go
vishnuitta: what if partitionStartByte is NOT divisible by LogicalBlockSize?

pkg/partition/partition.go
vishnuitta: ok.. so, endSector is included

pkg/partition/partition.go
akhilerm: partitionStartByte = 1MB = 1048576 bytes.

Hard drive vendors, have a fixed format for logical/physical sector sizes.  Current logical block sizes are 512 and 4k. 

@vishnuitta , are there cases of block sizes which are not multiples of 512 bytes. I am not sure of that.

References :
https://www.thomas-krenn.com/en/wiki/Advanced_Sector_Format_of_Block_Devices
[Western Digital](https://documents.westerndigital.com/content/dam/doc-library/en_us/assets/public/western-digital/collateral/white-paper/white-paper-advanced-format.pdf)
[Seagate] (https://www.seagate.com/in/en/tech-insights/advanced-format-4k-sector-hard-drives-master-ti/)
Logical Block size in [NVMe](https://metebalci.com/blog/a-quick-tour-of-nvm-express-nvme/#nvme-with-examples)

pkg/partition/partition_test.go
```diff
@@ -0,0 +1,128 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package partition
+
+import (
+	"github.com/diskfs/go-diskfs/partition/gpt"
```
vishnuitta, line 20: go mod doesn't seems to be having any 'version'.. is it better to use particular tag or version of go-diskfs?

pkg/partition/partition.go
```diff
@@ -0,0 +1,109 @@
...
+
+	table *gpt.Table
+}
+
+// ApplyPartitionTable applies the partition table to the given disk.
+func (d *Disk) ApplyPartitionTable() error {
+	if len(d.table.Partitions) == 0 {
+		return fmt.Errorf("no partitions specified in partition table")
+	}
+	diskFD, err := diskfs.Open(d.DevPath)
```
vishnuitta, line 56: Do we need to close the disk opened here?

cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -110,3 +171,51 @@ func (pe *ProbeEvent) initOrErrorEvent() {
...
+		// if available, other than the partition table UUID, because multiple partitions can have the same partition table
+		// UUID, but each partition will have a different UUID.
+		klog.Infof("device(%s) is a partition, using partition UUID: %s", bd.DevPath, bd.PartitionInfo.PartitionEntryUUID)
+		uuidField = bd.PartitionInfo.PartitionEntryUUID
+		ok = true
+	case len(bd.DeviceAttributes.WWN) > 0:
+		klog.Infof("device(%s) has a WWN, using WWN: %s", bd.DevPath, bd.DeviceAttributes.WWN)
+		uuidField = bd.DeviceAttributes.WWN
+		ok = true
+	case len(bd.FSInfo.FileSystemUUID) > 0:
```
pawanpraka1, line 209: what is the behavior on the Gke with local attached SSD? That will have file system, and later if we wipe that fs, will we deactivate that device?

cmd/ndm_daemonset/controller/probe.go
```diff
@@ -101,7 +101,6 @@ func (c *Controller) ListProbe() []*Probe {
 // FillBlockDeviceDetails lists registered probes and fills details from each probe
 func (c *Controller) FillBlockDeviceDetails(blockDevice *blockdevice.BlockDevice) {
 	blockDevice.NodeAttributes = c.NodeAttributes
-	blockDevice.DeviceAttributes.DeviceType = NDMDefaultDiskType
```
pawanpraka1, line 104: why do need to remove DeviceType? Pleae check if there is any component dependant on it.

pawanpraka1: nvm, I read it wrong.

pkg/partition/partition.go
vishnuitta: may not be.. but, just considering a case..
I vaguely remember disks of physical sectors of 4096 + few bytes to store parity..

pkg/partition/partition.go
akhilerm: The diskfs library does not have a close() method. I will check it and update in the upstream repo.

pkg/partition/partition_test.go
akhilerm: Will fix to a commit revision, when we move to go mod soon.

akhilerm:
> should we consider pushing an API to upstream which does all these things for a given device, i.e., an API that creates GPT table and a partition for entire disk.

@vishnuitta didn't get this comment. Which is the upstream you are mentioning? and what APIs should be pushed? `diskfs` already has the API to do these(create GPT table, write to disk etc)

cmd/ndm_daemonset/probe/eventhandler.go
akhilerm: BlockDevices with filesystem should be given to the user only when the VolumeMode in BDC is `FileSystem`. 
Still if the user wipes the filesystem, on next restart we will mark that block device as inactive. 
The restart is required because NDM currently only listens for udev events, which are not generated when filesystem is wiped.

blockdevice/blockdevice.go
```diff
@@ -323,3 +330,6 @@ const (
 	// claiming
 	Unclaimed string = "Unclaimed"
 )
+
+// Hierarchy is the block device hierarchy on the system
```
kmova, line 334: n_: this is a BlockDeviceCache which helps stores the discovered block devices in memory. This will help reduce the calls made to etcd for processing the disk events. The hierarchy information is secondary and is embedded into the block devices itself.

cmd/ndm_daemonset/controller/blockdevicestore.go
```diff
@@ -164,18 +164,21 @@ func (c *Controller) DeleteBlockDevice(name string) {
 		"ndm.blockdevice.delete.success", "Deleted blockdevice object ", name)
 }
 
-// ListBlockDeviceResource queries the etcd for the devices for the host/node
+// ListBlockDeviceResource queries the etcd for the devices
 // and returns list of blockdevice resources.
-func (c *Controller) ListBlockDeviceResource() (*apis.BlockDeviceList, error) {
+func (c *Controller) ListBlockDeviceResource(listAll bool) (*apis.BlockDeviceList, error) {
```
kmova, line 169: n_; add a comment. listAll=true implies that all the block devices in the cluster will be returned. and if listAll=false, the block device on the node where NDM is running will be returned.

cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -39,29 +46,49 @@ type ProbeEvent struct {
 
 // addBlockDeviceEvent fill block device details from different probes and push it to etcd
 func (pe *ProbeEvent) addBlockDeviceEvent(msg controller.EventMessage) {
-	bdList, err := pe.Controller.ListBlockDeviceResource()
+	bdAPIList, err := pe.Controller.ListBlockDeviceResource(true)
```
kmova, line 49: What is a `bdAPIList`?  is it `existingBDs` ?

cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -39,29 +46,49 @@ type ProbeEvent struct {
...
-		klog.Infof("Processed details for %s : %s", device.DevPath, device.UUID)
-		deviceInfo := pe.Controller.NewDeviceInfoFromBlockDevice(device)
+		klog.Infof("Processed details for %s", device.DevPath)
 
-		existingBlockDeviceResource := pe.Controller.GetExistingBlockDeviceResource(bdList, deviceInfo.UUID)
-		err := pe.Controller.PushBlockDeviceResource(existingBlockDeviceResource, deviceInfo)
-		if err != nil {
-			isErrorDuringUpdate = true
-			klog.Error(err)
+		// if GPTBasedUUID need to be used, generate the UUID,
```
kmova, line 69: These comments probably should now be moved into the `addBlockDevice` method, as the below code doesn't show the generation.

cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -72,32 +99,48 @@ func (pe *ProbeEvent) addBlockDeviceEvent(msg controller.EventMessage) {
...
+		// create the new UUID for removing the device
+		if isGPTBasedUUIDEnabled {
+			_, ok := pe.Controller.BDHierarchy[device.DevPath]
+			if !ok {
+				klog.Infof("Disk %s not in hierarchy", device.DevPath)
+				// not in hierarchy continue
+				continue
+			}
+			// remove from the hierarchy
+			delete(pe.Controller.BDHierarchy, device.DevPath)
```
kmova, line 120: the block device state -- is marked as inactive, but it is still present in the etcd. Will deleting from the cache possibly cause any issues with later processing. 

Can you explain what happens in the following case:

=> A detach of a device with partition occurs. Delete events will trigger. 
=> Attach event of the same device with partition occurs.

cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -72,32 +99,48 @@ func (pe *ProbeEvent) addBlockDeviceEvent(msg controller.EventMessage) {
...
+
+			uuid, ok := generateUUID(*device)
+			if !ok {
+				klog.Info("could not recreate UUID while removing device")
```
kmova, line 124: May be a comment on when such cases will occur will be good. 

This case can probably occur if a disk was being used for cstor (claimed) prior to upgrading to new uuid changes. And the disk is now removed from the node.

cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -72,32 +99,48 @@ func (pe *ProbeEvent) addBlockDeviceEvent(msg controller.EventMessage) {
...
+				continue
+			}
+			// use the new UUID for deactivating the blockdevice
+			device.UUID = uuid
+		}
+
+		existingBlockDeviceResource := pe.Controller.GetExistingBlockDeviceResource(bdAPIList, device.UUID)
```
kmova, line 131: Will this query fail for the claimed disks, new uuid generation is not done?  If a device is used by cstor, it might still get a new uuid because of the filesystem or the partitions.

cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -110,3 +153,254 @@ func (pe *ProbeEvent) initOrErrorEvent() {
...
+
+	if ok {
+		uuid = libudevwrapper.NDMBlockDevicePrefix + util.Hash(uuidField)
+		klog.Infof("generated uuid: %s for device: %s", uuid, bd.DevPath)
+	}
+
+	return uuid, ok
+}
+
+func (pe *ProbeEvent) addBlockDevice(bd blockdevice.BlockDevice) error {
```
kmova, line 205: Uff.. this is a very long function!!!

cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -110,3 +153,254 @@ func (pe *ProbeEvent) initOrErrorEvent() {
...
+		existingBlockDeviceResource := pe.Controller.GetExistingBlockDeviceResource(bdAPIList, deviceInfo.UUID)
+		err = pe.Controller.PushBlockDeviceResource(existingBlockDeviceResource, deviceInfo)
+		if err != nil {
+			klog.Errorf("creation of resource failed: %+v", err)
+			return err
+		}
+		return nil
+	} else {
+		klog.V(4).Infof("device: %s cannot be uniquely identified", bd.DevPath)
+		if len(bd.DependentDevices.Partitions) > 0 ||
```
kmova, line 387: q_: if partitions are present, uuid will be definitely generated right?

cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -110,3 +153,254 @@ func (pe *ProbeEvent) initOrErrorEvent() {
...
+						return err
+					}
+					return nil
+				}
+
+				klog.V(4).Infof("uuid: %s generated for parent device: %s", parentUUID, parentBD.DevPath)
+
+				parentBDAPI, err := pe.Controller.GetBlockDevice(parentUUID)
+
+				if errors.IsNotFound(err) {
```
kmova, line 290: Is this block of the code repeating for cases when generate uuid is not found or if the parent doesn't have a block device.

blockdevice/blockdevice.go
akhilerm: @kmova .
> this is a BlockDeviceCache which helps stores the discovered block devices in memory.

Yes. We use this map to keep track of blockdevices present on the host in memory. 

> This will help reduce the calls made to etcd for processing the disk events. The hierarchy information is secondary and is embedded into the block devices itself.

This will not help is reduce the calls to etcd. This cache / Hierarchy struct helps to get the details of dependent devices without performing a rescan on the system for all the disk attributes. 
eg: If at a certain point, we are processing `sda1` and we need to get the UUID of its parent device. We will get `/dev/sda` is the parent of `/dev/sda1`. This will be used to query the cache and get/generate the UUID of `/dev/sda`.

Will add the above info in the comment.

cmd/ndm_daemonset/controller/blockdevicestore.go
akhilerm: Done.

cmd/ndm_daemonset/probe/eventhandler.go
akhilerm: yes, the list of `BlockDevice` resources in the cluster.

cmd/ndm_daemonset/probe/eventhandler.go
akhilerm: Done.

cmd/ndm_daemonset/probe/eventhandler.go
akhilerm: > the block device state -- is marked as inactive, but it is still present in the etcd. Will deleting from the cache possibly cause any issues with later processing.

No. When the device gets reconnected to the node, it will be first added to the cache, and then marked as Active.

akhilerm: > => A detach of a device with partition occurs. Delete events will trigger.

Suppose the disk has 2 partitions `sda1` and `sda2`.
Sequence of events when device `sda` is disconnected
1. 3 udev `remove` events will be generated in the following order.
```    
remove sda1
remove sda2
remove sda
```
2. When remove event  for `sda1` and `sda2` is processed, we generate the UUID for `sda1` and `sda2`, and deactivate the resource. If the resource is not in etcd it is skipped. (for cstor created partitions it can be this case)
3. When remove event for `sda` is processed, we generate the UUID for `sda` and deactivate the resource.
NOTE: If UUID cannot be generated, we simple skip the event, because the add event gurantees that all devices will have a UUID (either by creating partition or by using WWN)


> => Attach event of the same device with partition occurs.

The above same disk is attached back. 3 udev `add` events will be generated in the following order
```
add sda
add sda1
add sda2
```
Steps for processing:
1. `udev add for sda`: We try to generate UUID for the device. If it fails, we check if there are partitions on the device, then we will skip the event.
2. `udev add for sda1/sda2`: generate UUID for the device. Check if device is claimed. If the device is claimed, update only the path, capacity, node attributes and state. If the device is not claimed, update all the details on the blockdevice resources.

cmd/ndm_daemonset/probe/eventhandler.go
akhilerm: Added.

No @kmova . This is the case when a remove event is generated for a device without any unique identifier. say a disk without WWN, we will just ignore it as there is no such blockdevice.

NOTE: This PR does not handle any cases of upgrade. It has to be done separately. All changes here are with the assumption this is a fresh installation.

cmd/ndm_daemonset/probe/eventhandler.go
akhilerm: A claimed device will always have UUID.

> If a device is used by cstor, it might still get a new uuid because of the filesystem or the partitions.

Can this case happen in fresh installations.?

NOTE: Device already used by cstor are not handled in this PR.

cmd/ndm_daemonset/probe/eventhandler.go
akhilerm: Yes. I will try to split it. Made it maximum verbose so that all cases are handled well.

cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -110,3 +153,254 @@ func (pe *ProbeEvent) initOrErrorEvent() {
...
+		deviceInfo := pe.Controller.NewDeviceInfoFromBlockDevice(&bd)
+		existingBlockDeviceResource := pe.Controller.GetExistingBlockDeviceResource(bdAPIList, deviceInfo.UUID)
+		err = pe.Controller.PushBlockDeviceResource(existingBlockDeviceResource, deviceInfo)
+		if err != nil {
+			klog.Errorf("creation of resource failed: %+v", err)
+			return err
+		}
+		return nil
+	} else {
+		klog.V(4).Infof("device: %s cannot be uniquely identified", bd.DevPath)
```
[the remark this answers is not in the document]
akhilerm: Done.

cmd/ndm_daemonset/probe/eventhandler.go
akhilerm: Yes, UUID will be generated. But whether resource will be created depends on whether the device has  any holders. Because if there are holders it means, we should not create for this device, but should create only for the holder device.

cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -110,3 +153,254 @@ func (pe *ProbeEvent) initOrErrorEvent() {
...
+			len(bd.DependentDevices.Holders) > 0 {
+			klog.V(4).Infof("device: %s has holders/partitions. %+v", bd.DevPath, bd.DependentDevices)
+		} else {
+			klog.Infof("starting to create partition on device: %s", bd.DevPath)
+			d := partition.Disk{
+				DevPath:          bd.DevPath,
+				DiskSize:         bd.Capacity.Storage,
+				LogicalBlockSize: uint64(bd.DeviceAttributes.LogicalBlockSize),
+			}
+			if err := d.CreateSinglePartition(); err != nil {
```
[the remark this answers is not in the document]
akhilerm: Yes. A single GPT partition that spans the entire disk will be created. When the partition table is written it will again trigger a udev add event for the partition. During processing this event, a UUID will be generated for the partition and BlockDevice resource will be created.

cmd/ndm_daemonset/probe/eventhandler.go
akhilerm: Yes @kmova . The code block is same. Just that log messages are different. Will see if I can combine them.

cmd/ndm_daemonset/probe/eventhandler.go
kmova: Could this be dependent on the kernel or is the order guaranteed?

cmd/ndm_daemonset/probe/eventhandler.go
akhilerm: refactored repeating code blocks into another method

cmd/ndm_daemonset/probe/eventhandler.go
akhilerm: Done.

PR #388, opened 2020-03-17
feat(apis): add block size and drive type fields

akhilerm:
add new fields into `spec.details` of blockdevice API
- logicalBlockSize
- physicalBlockSize
- hardwareSectorSize
- driveType

### DIFF
--- a/pkg/apis/openebs/v1alpha1/blockdevice_types.go
+++ b/pkg/apis/openebs/v1alpha1/blockdevice_types.go
@@ -123,8 +123,23 @@ type DeviceCapacity struct {
 
 // DeviceDetails represent certain hardware/static attributes of the block device
 type DeviceDetails struct {
-	// DeviceType represents the type of drive like SSD, HDD etc.,
+	// DeviceType represents the type of device like
+	// sparse, disk, partition, lvm, raid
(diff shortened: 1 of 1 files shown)


pkg/apis/openebs/v1alpha1/blockdevice_types.go
```diff
@@ -123,9 +123,24 @@ type DeviceCapacity struct {
 
 // DeviceDetails represent certain hardware/static attributes of the block device
 type DeviceDetails struct {
-	// DeviceType represents the type of drive like SSD, HDD etc.,
+	// DeviceType represents the type of device like
+	// sparse, disk, partition, lvm, raid
```
shovanmaity: Which probe populate this value?
Is there any chance we can get an empty value of it?

akhilerm: it can be filled in either by udevprobe or sysfs probe. No there will be no empty value in `deviceType`.

akhilerm:
This PR handles only the API changes, using the new fields in the API will be taken up in a separate PR and then both will be cherry-picked to v0.4.x branch

PR #394, opened 2020-03-20
fix(block-device): allow client to label and annotate bd

shovanmaity:
This PR fix metadata part of a block device. Now users can label and annotate block devices. What it does it actually takes metadata of older block device and patch that with system generated information.

### DIFF
--- a/cmd/ndm_daemonset/controller/blockdevicestore.go
+++ b/cmd/ndm_daemonset/controller/blockdevicestore.go
@@ -93,7 +93,42 @@ func (c *Controller) UpdateBlockDevice(blockDevice apis.BlockDevice, oldBlockDev
 		}
 	}
 
-	blockDeviceCopy.ObjectMeta.ResourceVersion = oldBlockDevice.ObjectMeta.ResourceVersion
+	// metadata of older object which contains -
+	// - name - no patch required we can use old object.
+	// - namespace - no patch required we can use old object.
+	// - generateName - no patch required we are not using it.
+	// - selfLink - populated by the system we should use old object.
+	// - uid - populated by the system we should use old object.
+	// - resourceVersion - populated by the system we should use old object.
+	// - generation - populated by the system we should use old object.
+	// - creationTimestamp - populated by the system we should use old object.
+	// - deletionTimestamp - populated by the system we should use old object.
+	// - deletionGracePeriodSeconds - populated by the system we should use old object.
+	// - labels - we will patch older labels with new labels.
+	// - annotations - we will patch older annotations with new annotations.
+	// - ownerReferences TODO
+	// - initializers TODO
+	// - finalizers TODO
+	// - clusterName - no patch required we can use old object.
+	metadata := oldBlockDevice.ObjectMeta
+
+	// Patch older label with new label. If there is a new key then it will be added
+	// if it is an existing key then value will be overwritten with value from new label
+	if blockDeviceCopy.ObjectMeta.Labels != nil {
+		for key, value := range blockDeviceCopy.ObjectMeta.Labels {
+			metadata.Labels[key] = value
+		}
+	}
+
+	// Patch older annotations with new annotations. If there is a new key then it will be added
+	// if it is an existing key then value will be overwritten with value from new annotations
+	if blockDeviceCopy.ObjectMeta.Annotations != nil {
+		for key, value := range blockDeviceCopy.ObjectMeta.Annotations {
+			metadata.Annotations[key] = value
+		}
+	}
+
+	blockDeviceCopy.ObjectMeta = metadata
 	blockDeviceCopy.Spec.ClaimRef = oldBlockDevice.Spec.ClaimRef
 	blockDeviceCopy.Status.ClaimState = oldBlockDevice.Status.ClaimState
 	err = c.Clientset.Update(context.TODO(), blockDeviceCopy)
--- a/cmd/ndm_daemonset/controller/blockdevicestore.go
+++ b/cmd/ndm_daemonset/controller/blockdevicestore.go
@@ -254,6 +254,8 @@ func (c *Controller) MarkBlockDeviceStatusToUnknown() {
 	}
 }
 
+// mergeMetadata merges oldMetadata with newMetadata. It takes old metadata and
+// update it's value with the help of new metadata.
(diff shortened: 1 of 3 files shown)


cmd/ndm_daemonset/controller/blockdevicestore.go
```diff
@@ -93,7 +93,42 @@ func (c *Controller) UpdateBlockDevice(blockDevice apis.BlockDevice, oldBlockDev
...
+	// - selfLink - populated by the system we should use old object.
+	// - uid - populated by the system we should use old object.
+	// - resourceVersion - populated by the system we should use old object.
+	// - generation - populated by the system we should use old object.
+	// - creationTimestamp - populated by the system we should use old object.
+	// - deletionTimestamp - populated by the system we should use old object.
+	// - deletionGracePeriodSeconds - populated by the system we should use old object.
+	// - labels - we will patch older labels with new labels.
+	// - annotations - we will patch older annotations with new annotations.
+	// - ownerReferences TODO
```
shovanmaity: for TODO items we need to decide should we allow the client to add them. If not then we can assign them with system generated value

cmd/ndm_daemonset/controller/blockdevicestore.go
```diff
@@ -93,7 +93,42 @@ func (c *Controller) UpdateBlockDevice(blockDevice apis.BlockDevice, oldBlockDev
...
+	// - initializers TODO
+	// - finalizers TODO
+	// - clusterName - no patch required we can use old object.
+	metadata := oldBlockDevice.ObjectMeta
+
+	// Patch older label with new label. If there is a new key then it will be added
+	// if it is an existing key then value will be overwritten with value from new label
+	if blockDeviceCopy.ObjectMeta.Labels != nil {
```
akhilerm: can this be made a separate private `func()`. I think the `Update` have now become too big to understand.

cmd/ndm_daemonset/controller/blockdevicestore_test.go
```diff
@@ -168,9 +168,11 @@ func TestUpdateDevice(t *testing.T) {
...
 	// Retrieve disk resource
 	cdevR, err := fakeController.GetBlockDevice(fakeDeviceUID)
-
-	devR.ObjectMeta.Name = "disk-updated-fake-uuid"
-	err = fakeController.UpdateBlockDevice(devR, cdevR)
+	if err != nil {
+		t.Fatal(err)
+	}
+	cdevR.ObjectMeta.Name = "disk-updated-fake-uuid"
+	err = fakeController.UpdateBlockDevice(*cdevR, nil)
```
akhilerm: Why is this change required?

shovanmaity: From the error log -
`If the resource is not present then it should return error.`
My understanding - there maybe 2 cases -
1. If the resource is not present we can get error while fetching it.
2. If the resource is not present we can get if we try to update it.

Earlier it was returning error for case 2 as was getting name and namespace from new object. Now we change that we are taking name and namespace from the older object so it will not fail.
Now it can fail for case 1 only and for that, we need to pass the old object as nil.

cmd/ndm_daemonset/controller/blockdevicestore.go
shovanmaity: I am ok with it.
Can you suggest what should that function do only merging metadata or full bd object

akhilerm: a function to patch objectMeta will be enough.
```
func mergeMetaData(oldBlockDevice.ObjectMeta, blockDeviceCopy.ObjectMeta) ObjectMeta {
   // All the todo comments
   // merging labels
   // merging annotations
   return metadata
}
```

cmd/ndm_daemonset/controller/blockdevicestore.go
```diff
@@ -253,3 +253,40 @@ func (c *Controller) MarkBlockDeviceStatusToUnknown() {
 		}
 	}
 }
+
```
akhilerm: can you provide the function comments.

cmd/ndm_daemonset/controller/blockdevicestore.go
```diff
@@ -253,3 +253,42 @@ func (c *Controller) MarkBlockDeviceStatusToUnknown() {
...
+	// - labels - we will patch older labels with new labels.
+	// - annotations - we will patch older annotations with new annotations.
+	// - ownerReferences as ndm-ds is not adding ownerReferences we can go with old object.
+	// - initializers ^^^
+	// - finalizers ^^^
+	// - clusterName - no patch required we can use old object.
+	metadata := oldMetadata
+	// Patch older label with new label. If there is a new key then it will be added
+	// if it is an existing key then value will be overwritten with value from new label
+	if newMetadata.Labels != nil {
```
sonasingh46: Do we need this nil check?

cmd/ndm_daemonset/controller/blockdevicestore.go
```diff
@@ -253,3 +253,42 @@ func (c *Controller) MarkBlockDeviceStatusToUnknown() {
...
+		for key, value := range newMetadata.Labels {
+			metadata.Labels[key] = value
+		}
+	}
+
+	// Patch older annotations with new annotations. If there is a new key then it will be added
+	// if it is an existing key then value will be overwritten with value from new annotations
+	if newMetadata.Annotations != nil {
```
sonasingh46: Similarly, do we need this nil check?

cmd/ndm_daemonset/controller/blockdevicestore.go
```diff
@@ -253,3 +253,42 @@ func (c *Controller) MarkBlockDeviceStatusToUnknown() {
...
+	// - creationTimestamp - populated by the system we should use old object.
+	// - deletionTimestamp - populated by the system we should use old object.
+	// - deletionGracePeriodSeconds - populated by the system we should use old object.
+	// - labels - we will patch older labels with new labels.
+	// - annotations - we will patch older annotations with new annotations.
+	// - ownerReferences as ndm-ds is not adding ownerReferences we can go with old object.
+	// - initializers ^^^
+	// - finalizers ^^^
+	// - clusterName - no patch required we can use old object.
+	metadata := oldMetadata
```
sonasingh46: Why we need a new `metadata` variable ? Can't we just use `oldMetadata`

PR #403, opened 2020-04-15
feat(partition): add pkg to create partitions

akhilerm:
add pkg to create partitions. The partition pkg is a wrapper over go-diskfs. Currently methods are added to create a single partition on the disk. Only GPT based partitions will be created on the disk.

### DIFF
--- a/pkg/partition/partition.go
+++ b/pkg/partition/partition.go
@@ -1,0 +1,151 @@
+/*
+Copyright 2020 The OpenEBS Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package partition
+
+import (
+	"fmt"
+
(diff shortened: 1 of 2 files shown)


akhilerm:
This PR contains the utility package related changes from #386

pkg/partition/partition.go
```diff
@@ -0,0 +1,111 @@
...
+	DevPath string
+	// DiskSize is size of disk in bytes
+	DiskSize uint64
+	// LogicalBlockSize is the block size of the disk normally 512 or 4k
+	LogicalBlockSize uint64
+
+	table *gpt.Table
+}
+
+// ApplyPartitionTable applies the partition table to the given disk.
```
kmova, line 53: Can we make these methods private to this function and just expose a single method that partitions the whole disk using the internal methods? Also, make this method idempotent so that if a partition exists - it doesn't recreate it. 

New methods can be added later that helps in creating additional partitions.

akhilerm: Why do we have to do in such a way @kmova ? 
That will be abstracting too much details right? we can make the function to error out if a partition already exists.

akhilerm: We can create a method over it , instead of calling 3 separate methods. But still I am not in the strong opinion of making a single methods public and rest of them private.

kmova: How do we ensure that partitioning gets called only once and have the validation that partition should be created only if the device doesn't already have one?

We can leave it up to the caller, but what if there are multiple callers - possibly due to a bug in the future contributions. 

It helps to lock down as much as possible so that callers don't make mistakes.

akhilerm: Done.

PR #407, opened 2020-04-23
fix(controller): fix using namespace in controller

akhilerm:
- initialize namespace before setting on manager
- fix using namespace in Get() call

### DIFF
--- a/cmd/ndm_daemonset/controller/blockdevicestore.go
+++ b/cmd/ndm_daemonset/controller/blockdevicestore.go
@@ -32,6 +32,9 @@ import (
 // blockDevice is DeviceResource-CR
 func (c *Controller) CreateBlockDevice(blockDevice apis.BlockDevice) error {
 
+	// set namespace on the api resource
+	blockDevice.SetNamespace(c.Namespace)
+
 	blockDeviceCopy := blockDevice.DeepCopy()
 	err := c.Clientset.Create(context.TODO(), blockDeviceCopy)
 	if err == nil {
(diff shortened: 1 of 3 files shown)


cmd/ndm_daemonset/controller/blockdevicestore.go
```diff
@@ -32,6 +32,9 @@ import (
 // blockDevice is DeviceResource-CR
 func (c *Controller) CreateBlockDevice(blockDevice apis.BlockDevice) error {
 
+	// set namespace on the api resource
```
kmova, line 35: It is good that this is being set now. Prior to this change also, BDs were namespace scoped, was K8s automatically setting them?

akhilerm: earlier we were setting the namespace while converting to the BD API resource. Now we are setting it just before creating. So that the namespace of client is provided. 

Also, we were previously not using the `GetBlockDevice()` method anywhere in our code. In that we were not using Namespace, it was given as empty. So it will cause the Get() call to error out. This was found out while working with the other PR.

cmd/ndm_daemonset/controller/controller.go
```diff
@@ -131,23 +130,24 @@ func NewController() (*Controller, error) {
...
+	// get the namespace in which NDM is installed
+	ns, err := getNamespace()
 	if err != nil {
 		return controller, err
 	}
+	controller.Namespace = ns
 
-	// Setup Scheme for all resources
-	if err := apis.AddToScheme(mgr.GetScheme()); err != nil {
+	mgr, err := manager.New(controller.config, manager.Options{Namespace: controller.Namespace})
```
akhilerm, line 140: @kmova , Also here we were setting empty namespace on the manager. But since the resource had namespace field in them, it worked correctly.

PR #409, opened 2020-04-28
fix(filter): add default vendor OpenEBS in vendor filter

akhilerm:

This PR adds `OpenEBS` to the list of default excluded vendors. This is done so that, even if user removes the vendor filter, OpenEBS created iscsi disks are excluded by NDM.

**What this PR does?**:
- add OpenEBS as a default vendor in vendor filter
- modify test cases to add OpenEBS as a default vendor

No

None

**Checklist:**
- [ ] Fixes #<issue number>
- [x] Commit has unit tests

### DIFF
--- a/cmd/ndm_daemonset/filter/vendorfilter.go
+++ b/cmd/ndm_daemonset/filter/vendorfilter.go
@@ -26,6 +26,8 @@ import (
 
 const (
 	vendorFilterKey = "vendor-filter"
+	// vendorValueOpenEBS is the vendor for iscsi disks created by OpenEBS
+	vendorValueOpenEBS = "OpenEBS"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


cmd/ndm_daemonset/filter/vendorfilter.go
```diff
@@ -26,13 +26,18 @@ import (
 
 const (
 	vendorFilterKey = "vendor-filter"
+	// OpenEBSVendor is the vendor for iscsi disks created by OpenEBS
+	OpenEBSVendor = "OpenEBS"
```
kmova, line 30: n_: May be useful to change the const variable name to "VendorValueOpenEBS"? 

also, check if go recommends any specific representation for const values. So that while they are being used in the code below, the reader can know if it is constant by looking at the naming convention.

akhilerm: Go recommends using CamelCase for constants, as is in the standard library. 
I will rename it to `vendorValueOpenEBS`, so that it is un-exported .  We need to use it only in this package.

akhilerm: Again, a thought why should we have the word `value` in it. 
It sounds like naming a literal `NameString`, `AddressHashMap`,  which is not recommended.

akhilerm: Done.

PR #426, opened 2020-05-22
fix(featuregate): make feature gate independent of controller

akhilerm:

The feature gate used to be a part of the daemon controller. This restricted its use to only code paths which had access to the controller pointer. This PR removes the dependency of feature gates on the daemon controller and can be used globally across the code base.

**What this PR does?**:
- adds a new exporter variable `FeatureGates` in features package which holds the state of all features.
- remove feature gate from daemon controller

No

- tested by providing a set of correct and incorrect feature gates.

**Checklist:**
- [ ] Fixes #<issue number>
- [x] Commit has unit tests

### DIFF
--- a/pkg/features/features.go
+++ b/pkg/features/features.go
@@ -13,6 +13,14 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
+/*
+	The contents of this package has its origins from the feature gate
+	implementation in kubernetes.
+	Refer :
+		https://github.com/kubernetes/component-base/tree/master/featuregate
+		https://github.com/kubernetes/kubernetes/tree/master/pkg/features
+
+*/
 
 package features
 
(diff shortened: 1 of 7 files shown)


pkg/features/features.go
```diff
@@ -38,22 +43,45 @@ const (
...
+var defaultFeatureGates = map[Feature]bool{
+	GPTBasedUUID: false,
+}
+
+// featureGate is a type that implements the FeatureGate interface
+type FeatureGate map[Feature]bool
+
+// DefaultFeatureGate is the global feature gate that can be used to check if a feature is enabled
+// or disabled
+var DefaultFeatureGates = NewFeatureGate()
```
kmova, line 62: n_: DefaultFeatureGates => FeatureGates or a generic name as it contains both default and those enabled via ENV etc,. right?

akhilerm: `FeatureGates` was not used because there is already a `FeatureGate` type in the same package.

Do you have suggestion for any other generic names?

akhilerm: `FeatureGate` will renamed to `featureFlag`, and `DefaultFeatureGates` to `FeatureGates`

PR #429, opened 2020-06-01
feat(dependency): Migrated to Go modules

harshthakur9030:
This migrates the repo to use Go modules which is a better dependency management for Golang. 

**What this PR does?**:
Migrates to Go module

Goinkgo and Gomega were upgraded 

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

.


Makefile
```diff
@@ -251,7 +250,8 @@ docker.exporter: build.exporter Dockerfile.exporter
 .PHONY: deps
 deps: header
 	@echo '--> Resolving dependencies...'
-	dep ensure
+	go mod tidy 
+	go mod download
```
akhilerm, line 254: If we do a `go mod download`, the dependencies will be downloaded to `$GOPATH/pkg/mod`. But for building we will still be using the vendor directory right? Can we fix that.

PR #432, opened 2020-06-02
feat(dependency): Migrated to go modules

harshthakur9030:
Go modules is a better dependency management than dep.

**What this PR does?**:
Migrates to use Go modules

 Goinkgo and Gomega were upgraded

Works on go 1.13 and 1.14

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests


Makefile
```diff
@@ -162,21 +160,14 @@ vet:
...
 # shellcheck target for checking shell scripts linting
 .PHONY: shellcheck
 shellcheck: getshellcheck
 	find . -type f -name "*.sh" | grep -v "./vendor/*" | xargs /tmp/shellcheck-stable/shellcheck
 
 .PHONY: getshellcheck
 getshellcheck:
-	wget -c 'https://github.com/koalaman/shellcheck/releases/download/stable/shellcheck-stable.linux.x86_64.tar.xz' --no-check-certificate -O - | tar -xvJ -C /tmp/
-
+ 	wget -c 'https://github.com/koalaman/shellcheck/releases/download/stable/shellcheck-stable.linux.x86_64.tar.xz' --no-check-certificate -O - | tar -xvJ -C /tmp/
```
akhilerm, line 170: This is not required right? The change is already present.

harshthakur9030: The previous build failed for shell check, I assumed I forgot to add the URL fix but even now it fails. I can git reset and hard push once we figure out what's causing it

akhilerm: The error says not able to find the shellcheck binary in the path

harshthakur9030: It finds it in other architectures.

akhilerm: shellcheck is enabled only in `amd64`, thats why its not failing in other architectures.

harshthakur9030: Oh. How do we fix it?

akhilerm:
The changes looks good, Tried building and it all works fine. Can you also check if we need to pass `-mod` argument to `gox` command while building?

Makefile
```diff
@@ -251,7 +244,8 @@ docker.exporter: build.exporter Dockerfile.exporter
 .PHONY: deps
 deps: header
 	@echo '--> Resolving dependencies...'
-	dep ensure
+	go mod vendor
+	go mod tidy
```
mynktl, line 248: sequence can be wrong. tidy updates the module list and vendor will make copy of dependency. Better to run `go mod tidy` first then `go mod vendor`

mynktl: You may need to check go.mod/go.sum to verify dependency. Check  [this example](https://github.com/openebs/velero-plugin/blob/master/Makefile#L90)  for verification.

mynktl:
@harshthakur9030 You also need to remove `Gopkg.lock` and `Gopkg.toml` file since these files are not required anymore.

harshthakur9030:
Done! @mynktl

harshthakur9030:
I just noticed the integration test failed again.

harshthakur9030:
Before merging, we might need to fix the weird error for integration test. No clue why that's happening @akhilerm @mynktl

akhilerm:
> Before merging, we might need to fix the weird error for integration test. No clue why that's happening @akhilerm @mynktl

I am trying to reproduce it locally, But with no result.

akhilerm:
@harshthakur9030 , if you look at the log messages [here](https://travis-ci.org/github/openebs/node-disk-manager/jobs/694892781#L1092), you can see that while running the integration test. the modules are not fetched. When tried locally, first it fetched the modules and then ran the integration test. Not sure if it is the exact cause. you may need to check on it.

harshthakur9030:
Oh, can you compare your go env with Travis?

harshthakur9030:
Closing this. I will send a new PR with clean commit history

PR #433, opened 2020-06-05
feat(api): adding an API to expose functionality of NDM

harshthakur9030:
It adds a new gRPC API. 

**What this PR does?**:
Adds a gRPC API which can be called to perform the following operations:
*  Version of NDM
*  Node name 
*  Lists block devices and their relationships
*  SMART stats for a block devices
*  Sets and gets Hugepages on a node
*  Rescan method to sync NDM with etcd

No

    * iSCSI status has been tested on GKE, Digital Ocean Kubernetes and CentOS image. 
    * SMART stats feature doesn't have unit/integration tests as there's no clean way to test it. It has been manually verified

    * Setting Hugepages through the gRPC method doesn't guarantee that demanaded pages will be set because of availablity of memory.

**Checklist:**
- [ ] Fixes #<issue number>
- [x] Commit has unit tests
- [x] Commit has integration tests


cmd/ndm-grpc/main.go
```diff
@@ -0,0 +1,64 @@
...
+	protos "github.com/openebs/node-disk-manager/pkg/ndm-grpc/protos/ndm"
+	"github.com/openebs/node-disk-manager/pkg/ndm-grpc/server/services"
+	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/reflection"
+)
+
+func main() {
+
+	log := logrus.New()
```
akhilerm, line 29: can we use klog for logging?

cmd/ndm-grpc/main.go
```diff
@@ -0,0 +1,64 @@
...
+
+	// Similar to registring handlers for http
+	protos.RegisterInfoServer(gs, is)
+
+	protos.RegisterISCSIServer(gs, ss)
+
+	protos.RegisterNodeServer(gs, ns)
+
+	l, err := net.Listen("tcp", "0.0.0.0:9090")
+	if err != nil {
```
akhilerm, line 55: This whole code will be moved to the daemon right?

cmd/ndm_daemonset/app/command/start.go
```diff
@@ -28,6 +38,49 @@ import (
 
 //NewCmdStart starts the ndm controller
 func NewCmdStart() *cobra.Command {
+	var wg sync.WaitGroup
+
+	go func() {
```
akhilerm, line 43: Ignore the previous comment, This routine should be started just after we start the controller.

dev.yaml
```diff
@@ -0,0 +1,351 @@
+# Create NDM configmap
```
akhilerm, line 1: Try to add changes to ndm-operator.yaml, instead of creating a new  yaml file.

harshthakur9030: This is something I'm using during development,  I've attached another container to the daemonset which makes it easy to test the API. I need to make some changes to the Makefile as well. I'll remove all of this(build process) once the code looks good.

pkg/ndm-grpc/server/services/listBlockDevice.go
```diff
@@ -0,0 +1,192 @@
...
+		return nil, status.Errorf(codes.Internal, "Error fetching list of disks")
+	}
+
+	if len(blockDeviceList.Items) == 0 {
+		n.Log.Info("No items found")
+	}
+
+	blockDevices := make([]*protos.BlockDevice, 0)
+
+	parentNames, err := GetParentDisks(n, blockDeviceList)
```
akhilerm, line 73: Can a comment be provided here, what we are trying to. I am not able to understand the output format.

pkg/ndm-grpc/server/services/listBlockDevice.go
```diff
@@ -0,0 +1,192 @@
...
+	}, nil
+}
+
+// GetHolders gets the holders of a particular device
+func GetHolders(n *Node, BL *v1alpha1.BlockDeviceList) ([]string, error) {
+
+	blockDeviceNames := make([]string, 0)
+	for _, bd := range BL.Items {
+		device := hierarchy.Device{Path: bd.Spec.Path}
+		depDevices, err := device.GetDependents()
```
akhilerm, line 125: try limiting the call to `device.GetDependents`, It actually parses sysfs.

For each device in list you can call `GetDependents` only once and then get all the dependent devices, which can be further updated into the `Node` to get the required result.

cmd/ndm-grpc/main.go
harshthakur9030: Done

cmd/ndm_daemonset/app/command/start.go
```diff
@@ -54,8 +66,48 @@ func NewCmdStart() *cobra.Command {
 			// Start starts registering of probes present in RegisteredProbes
 			probe.Start(probe.RegisteredProbes)
 			ctrl.Start()
+
 		},
 	}
+	go func() {
+		{
+			wg.Add(1)
```
akhilerm, line 74: Why do we require a `WaitGroup` here?

go.mod
```diff
@@ -27,18 +31,21 @@ require (
 	github.com/pborman/uuid v0.0.0-20180906182336-adf5a7427709 // indirect
 	github.com/peterbourgon/diskv v2.0.1+incompatible // indirect
 	github.com/prometheus/client_golang v0.9.2
-	github.com/prometheus/client_model v0.0.0-20190129233127-fd36f4220a90 // indirect
 	github.com/prometheus/common v0.2.0 // indirect
 	github.com/prometheus/procfs v0.0.0-20190227231451-bbced9601137 // indirect
+	github.com/sirupsen/logrus v1.5.0 // indirect
```
akhilerm, line 36: We are not using logrus. This can be removed. Also the checked in logrus files in vendor.

pkg/ndm-grpc/protos/ndm.proto
```diff
@@ -0,0 +1,101 @@
...
+  // Rescan syncs etcd and NDM's local state
+  rpc Rescan(Null) returns (Message);
+}
+
+message Message{
+  string msg = 1;
+}
+
+message Hugepages{
+  int32 pages = 1;
```
akhilerm, line 48: I think we have both `2Mi` and `1Gi` pages. The name should specify which page we are returning.

pkg/ndm-grpc/server/services/listBlockDevice.go
```diff
@@ -0,0 +1,281 @@
...
+}
+
+type disks []protos.BlockDevice
+
+//AllBlockDevices contains all the relationships and device types
+type AllBlockDevices struct {
+	Parents    []string
+	Partitions []string
+	LVMs       []string
+	RAIDs      []string
```
akhilerm, line 51: LVM, RAID, Loop and Sparse are not relationship between devices. Its a deviceType.

pkg/ndm-grpc/server/services/listBlockDevice.go
```diff
@@ -0,0 +1,281 @@
...
+	if err != nil {
+		klog.Errorf("Error fetching Parent disks %v", err)
+	}
+
+	for _, name := range all.Parents {
+
+		blockDevices = append(blockDevices, &protos.BlockDevice{
+			Name:       name,
+			Type:       "Disk",
+			Partitions: FilterParitions(name, all.Partitions),
```
akhilerm, line 98: Why are we taking all partitions and then filtering it? We can get Partitions for single device right?

pkg/ndm-grpc/protos/ndm.proto
harshthakur9030: Will a comment do? Since we're only using 2Mi pages?

go.mod
harshthakur9030: Haven't imported it directly anywhere . It's being used by some other packages I imported so we cannot remove it.

pkg/ndm-grpc/server/services/listBlockDevice.go
harshthakur9030: The struct is supposed to have all the relationships and device types. Please suggest a better name for it.

pkg/ndm-grpc/server/services/listBlockDevice.go
harshthakur9030: It is to minimize the calls made to `GetDependents`. All the partitions found on a node are stored in  `all.Partitions`.   `GetDependents` is called on each block device only when necessary. Later, while sending the response, we can just filter accordingly based on the block device.

akhilerm:
@harshthakur9030  Can we make the grpc API available via the feature flag. once its stabilized we will make it default. You can refer [here](https://github.com/openebs/node-disk-manager/blob/3457e6a484d010e2dac649cafd6a88956599d6ec/pkg/features/features.go#L39)

pkg/features/features.go
```diff
@@ -41,23 +41,27 @@ const (
 	// blockdevice UUID algorithm mentioned in
 	// https://github.com/openebs/openebs/pull/2666
 	GPTBasedUUID Feature = "GPTBasedUUID"
+	// GRPC feature flag starts the GRPC server which provides functionality to manage block devices
+	GRPC Feature = "GRPC"
```
akhilerm, line 45: Better to use `GRPCServer` rather than just grpc.

harshthakur9030: Done

ndm-operator.yaml
```diff
@@ -121,13 +121,16 @@ spec:
       # Use host network as container network to monitor udev source using netlink
       # to detect disk attach and detach events using fd.
       hostNetwork: true
+      hostPID: true
```
akhilerm, line 124: @kmova Will it be an issue to use `hostPID` here.
To check iscsi status, it is required.

kmova:
@harshthakur9030 a few suggestions, can we take the following non-grpc tasks into their own PRs: 
- Removing the dependency on gox and migrating to go-build
- bumping-up of go version
- refactor the code to introduce te `re-scan` capability

Also, as part of this PR, add a developer doc to specify how to introduce new API

harshthakur9030:
https://github.com/openebs/node-disk-manager/pull/455 has the rescan PR and https://github.com/openebs/node-disk-manager/pull/456 removes gox. Bumping go version doesn't impact vendor files, grpc api has imported a lot of new packages, will try to seperate it in commits once the above are merged.

cmd/ndm_daemonset/app/command/start.go
```diff
@@ -54,8 +70,49 @@ func NewCmdStart() *cobra.Command {
...
+			os.Exit(1)
+		}
+
+		// Listen for requests
+		klog.Info("Starting server at 9090")
+		gs.Serve(l)
+
+	}
+
+}
```
akhilerm, line 118: Thinking out loud. This function can be made to a separate package so that we can keep the `command` package clean.

The new package can be `cmd/ndm_daemonset/grpcserver`

harshthakur9030: Makes sense.

harshthakur9030: Method can be Start() . So from this file, we call grpcserver.Start()

pkg/features/features.go
```diff
@@ -41,23 +41,27 @@ const (
 	// blockdevice UUID algorithm mentioned in
 	// https://github.com/openebs/openebs/pull/2666
 	GPTBasedUUID Feature = "GPTBasedUUID"
+	// GRPC feature flag starts the GRPC server which provides functionality to manage block devices
+	GRPCServer Feature = "GRPCServer"
```
kmova, line 45: rename GRPCServer to "APIService"

pkg/ndm-grpc/protos/ndm.proto
```diff
@@ -0,0 +1,103 @@
+syntax = "proto3";
```
kmova, line 1: push ndm.proto to root directory.

kmova: push the ndm.pb.go to root directory "spec/ndm/ndm.pb.go"

pkg/ndm-grpc/server/info.go
```diff
@@ -0,0 +1,18 @@
+/*
```
kmova, line 1: maybe good to move this to root directory "api-service" 

The content under "server" may be moved to "api-service/node" - indicating these are the services offered by per node (daemonset). 

Going forward the cluster level services can be added under "api-service/cluster".

harshthakur9030: About the naming convention, underscores and spaces aren't allowed so that leaves us with only hyphens. Couldn't find anything about cases though.

akhilerm:
@harshthakur9030  Can you rebase this PR, there has been some changes to the vendor and module files.

go.mod
```diff
@@ -14,6 +16,8 @@ require (
 	github.com/stretchr/testify v1.5.1
 	golang.org/x/sync v0.0.0-20190911185100-cd5d95a43a6e
 	golang.org/x/sys v0.0.0-20200519105757-fe76b779f299
+	google.golang.org/grpc v1.30.0
+	google.golang.org/protobuf v1.23.0
```
akhilerm, line 20: Why is the protobuf imported twice?

pkg/util/util.go
```diff
@@ -120,3 +120,25 @@ func StateStatus(state bool) string {
 	}
 	return status[state]
 }
+
+// AddUniqueStringtoSlice ensures there are no repeated devices added to a slice
+func AddUniqueStringtoSlice(names []string, name string) []string {
```
akhilerm, line 125: this should be in `pkg/util/strutil.go`

go.mod
harshthakur9030: They're both different packages. Grpc is the framework used and protobuf is how data serialization is done. Similar to SOAP and XML.

harshthakur9030: So both the packages have been imported by the auto generated code which we don't control. One is being used for the proto files and other for the protocol itself. At some point, the compiler will start using google.golang.org/protobuf , I suppose as I read this: https://developers.google.com/protocol-buffers/docs/reference/go/faq

api-service/node/services/listBlockDevice.go
```diff
@@ -0,0 +1,257 @@
...
+	"github.com/openebs/node-disk-manager/pkg/hierarchy"
+	"github.com/openebs/node-disk-manager/pkg/util"
+	protos "github.com/openebs/node-disk-manager/spec/ndm"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"k8s.io/klog"
+)
+
+// Node helps in using types defined in Server
```
akhilerm, line 32: What is the `Server` here?

api-service/node/services/listBlockDevice.go
```diff
@@ -0,0 +1,257 @@
...
+func (n *Node) ListBlockDevices(ctx context.Context, null *protos.Null) (*protos.BlockDevices, error) {
+	klog.Info("Listing block devices")
+
+	ctrl, err := controller.NewController()
+	if err != nil {
+		klog.Errorf("Error creating a controller %v", err)
+		return nil, status.Errorf(codes.NotFound, "Namespace not found")
+	}
+
+	err = ctrl.SetControllerOptions(controller.NDMOptions{ConfigFilePath: "/host/node-disk-manager.config"})
```
akhilerm, line 68: Make this path also into constant. 

Later we may need to make this configurable as done in the controller.

ndm.proto
```diff
@@ -0,0 +1,112 @@
...
+// This service can return generic information 
+service Info {
+  // FindVersion Method gets the version information of the NDM components installed
+  rpc FindVersion (Null) returns (VersionInfo);
+  }
+
+service Node{
+      
+  // FindNodeName method is used find the name of the node on which NDM is running on 
+  rpc Name(Null) returns (NodeName) ;
```
akhilerm, line 30:
```suggestion
  rpc Name(Null) returns (NodeName);
```

akhilerm: Also the comment, can you check it.

api-service/node/services/listBlockDevice.go
```diff
@@ -0,0 +1,257 @@
...
+
+	}
+	klog.V(4).Infof("Parent Devices found are: %v", ParentDeviceNames)
+	klog.V(4).Infof("Partitions  found are: %v", PartitionNames)
+
+	klog.V(4).Infof("LVM found are: %v", LVMNames)
+	klog.V(4).Infof("RAID disks found are: %v", RAIDNames)
+
+	klog.V(4).Infof("Holder Devices found are: %v", HolderDeviceNames)
+	klog.V(4).Infof(" Slaves found are: %v", SlaveDeviceNames)
```
akhilerm, line 222:
```suggestion
	klog.V(4).Infof("Slaves found are: %v", SlaveDeviceNames)
```

ndm.proto
```diff
@@ -0,0 +1,112 @@
...
+  
+  // ListBlockDevices method is used to return all the block devices found in hierarchy by NDM
+  rpc ListBlockDevices(Null) returns (BlockDevices);
+  
+  // This method returns status of ISCSI
+  rpc ISCSIStatus(Null) returns (Status);
```
akhilerm, line 36: Getting an error while checking iSCSI status.
```
Error invoking method "ndm.Node.ISCSIStatus": service "ndm.Node" does not include a method named "ISCSIStatus"
```

ndm.proto
```diff
@@ -0,0 +1,114 @@
...
+  
+  // ISCSIStatus method returns status of ISCSI
+  rpc ISCSIStatus(Null) returns (Status);
+
+  // ListBlockDeviceDetails method is used to find SMART info about a block device.
+  // Only the name field is required for input 
+  rpc ListBlockDeviceDetails(BlockDevice) returns (BlockDeviceDetails); 
+
+  // SetHugepages sets 2MB hugePages on this node.  
+  // Note: Calling this method doesn't gurantee that the said number of pages will be set.
```
akhilerm, line 45:
```suggestion
  // Note: Calling this method doesn't guarantee that the said number of pages will be set.
```

PR #442, opened 2020-06-18
feat(upgrade, api, core): upgrade old BlockDevices to new BlockDevices with new UUID

akhilerm:

The BDs existing in old UUID scheme need to be migrated to BDs in new scheme.
BlockDevices will be creates for disks that are being used by zfs localPV and mayastor.

**What this PR does?**:
Handles cases for co-exisiting of BDs in legacy and GPTBased uuid scheme
Cases which are handled:
1. All BlockDevices that are not in `Unclaimed` (`Claimed` and `Released`) state and uses the legacy uuid scheme will be annotated with `internal.openebs.io/uuid-scheme=legacy`
2. BlockDevices that are in `Unclaimed` state will be made inactive and new BlockDevice resource using the GPT scheme will be created.
3. Devices that are newly connected (first time appearing in the cluster) will have BlockDevice resources created using the GPT scheme.
4. Virtual Devices (which donot have a unique identifier) which are currently in Claimed state, when disconnected and reconnected at a different path , new BlockDevice resource will be created with the legacy uuid scheme and the `internal.openebs.io/uuid-scheme=legacy` annotation. This is done so as to make the changes backward compatible with applications that are currently using those BlockDevice resources.

BlockDevice resource will be created for devices that are used by ZFS-LocalPV. Those blockdevices will be labelled `openebs.io/block-device-tag=zfs-localpv`  so that the operator cannot claim those resources. 

MayaStor disks: Mayastor disks will be ignored, because we need a unique way to identify disks that are already in use by mayastor. (This is because mayastor disks does not have any partitions.)

Pre-requisites before performing the upgrade:
1. all the devices that are in use should be claimed (i.e There are chances that multiple block device resource can exist if the device path has changed. It should be made sure that the actual BD resource that corresponds to the device in use should be in claimed state. Commonly occurs in cstor, because zfs scan the /dev directory and finds the disk even if path has changed.)
2. all data disks that is being used by the storage engine should be connected to the node and `1.` should be satisfied
3. all manually created and claimed blockdevices should excluded in the configmap

No.

- [x] LocalPV created on virtual disk -> Upgrade NDM -> Resource updated with annotations. (Restarts also handled)
- [x] LocalPV created on virtual disk -> disconnected the disk -> disk comes at a different path. In this case, the device wont be used by local PV anymore since the mountpath has changed and pod cannot mount the device. (same case for moving to another node also.) After upgrade a new resource will be created using fs uuid. If path does not change and pod mounts the device then works as expected by adding the annotation.
- [x] LocalPV created on physical disk -> upgrade NDM -> Resource updated with annotations (Restart also will be similar case)
- [x] cstor on virtual disk -> upgrade NDM -> move virtual disk to another node
- [x] cstor on virtual disk -> upgrade NDM -> disconnect and reconnect the device
- [x] cstor on physical disk -> upgrade NDM -> move disk to different node (also disconnect and reconnect the device)

NOTES:
- After a legacy resource is unclaimed, a rescan/restart of pod would have to performed so as to keep the data in sync
- Does not support migration of localPV raw block.

Flow Diagram for the upgrade of LocalPV is available [here](https://gist.github.com/akhilerm/cb9cfff929bff00efaa0d3f70d25a116#file-localpvfsupgrade-pdf). The gist also contains the changes that are required in the steps for cstor.

**Checklist:**
- [x] Fixes [#3043](https://github.com/openebs/openebs/issues/3043)
- [x] Commit has unit tests
- [x] (Optional) Are upgrade changes included in this PR? If not, mention the issue/PR to track: 


cmd/ndm_daemonset/probe/deleteDeviceHandler.go
```diff
@@ -0,0 +1,52 @@
...
+
+	// try to deactivate resource with both UUIDs. This is required in the following case
+
+	existingLegacyBlockDeviceResource := pe.Controller.GetExistingBlockDeviceResource(bdAPIList, bd.UUID)
+	if existingLegacyBlockDeviceResource != nil {
+		pe.Controller.DeactivateBlockDevice(*existingLegacyBlockDeviceResource)
+		klog.V(4).Infof("deactivated device: %s, using legacy UUID", bd.DevPath)
+	}
+
+	existingBlockDeviceResource := pe.Controller.GetExistingBlockDeviceResource(bdAPIList, uuid)
```
pawanpraka1, line 45: what if generateUUID fails? should we do this if uuid is valid?

cmd/ndm_daemonset/probe/deleteDeviceHandler.go
```diff
@@ -0,0 +1,52 @@
...
+	// legacy UUID will be generated at event processing and filled into blockdevice UUID field, no need to regerate here.
+	klog.V(4).Infof("device: %s disconnected, generated legacy UUID: %s.", bd.DevPath, bd.UUID)
+
+	uuid, ok := generateUUID(bd)
+	// this can happen if the device didn't have any unique identifiers
+	if !ok {
+		klog.Info("could not recreate GPT-UUID while removing device")
+	}
+
+	// try to deactivate resource with both UUIDs. This is required in the following case
```
pawanpraka1, line 37: why we are trying both the method, gpt and legacy? We should try one of them right?

ndm-operator.yaml
```diff
@@ -142,6 +142,8 @@ spec:
         - name: procmount
           mountPath: /host/proc
           readOnly: true
+        - name: devmount
```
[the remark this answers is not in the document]
akhilerm: yes, 
For handling cases where the device is with the pool is connected after the NDM pod has started. i.e the pool is created on a device which was claimed using the legacy UUID. The following sequence of events will happen
1. NDM 0.6 is used and user creates a pool on `/dev/sdb` (blockdevice-abc which uses legacy UUID)
2. User disconnects the `/dev/sdb`
3. Upgrade to the new version of NDM which support partitions
4. Connects the drive, assume it came back at `/dev/sdb`. In this case we should be able to read the filesystem on the newly connected disk. Then only we can understand that the device is currently in use. 
Otherwise NDM will try to create resources for the partitions `sdb1` and `sdb9`  which should not happen.

If we are able to access and read the disk, then we can identify that disk is already in use. 

@pawanpraka1 You see any issues with mounting this. Similar to the bug that was hit with open FDs.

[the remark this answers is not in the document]
kmova: To make it clear, can we add comments to each of the mount points indicating why we need them.

cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -109,34 +109,18 @@ func (pe *ProbeEvent) deleteBlockDeviceEvent(msg controller.EventMessage) {
 	isGPTBasedUUIDEnabled := features.FeatureGates.IsEnabled(features.GPTBasedUUID)
 
 	for _, device := range msg.Devices {
-		// create the new UUID for removing the device
+		// create the new UUID and legacy UUID for removing the device
```
kmova, line 112: Can we expand on this comment a little more. Unable to understand the reason for refactoring by just looking the changes in this file.

cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -109,34 +109,18 @@ func (pe *ProbeEvent) deleteBlockDeviceEvent(msg controller.EventMessage) {
...
-				continue
-			}
-			// remove from the hierarchy
-			delete(pe.Controller.BDHierarchy, device.DevPath)
-
-			uuid, ok := generateUUID(*device)
-			// this can happen if the device didn't have any unique identifiers
-			if !ok {
-				klog.Info("could not recreate UUID while removing device")
+			_ = pe.deleteBlockDevice(*device, bdAPIList)
```
kmova, line 114: :+1:

cmd/ndm_daemonset/probe/deleteDeviceHandler.go
```diff
@@ -0,0 +1,52 @@
...
+
+package probe
+
+import (
+	"github.com/openebs/node-disk-manager/blockdevice"
+	apis "github.com/openebs/node-disk-manager/pkg/apis/openebs/v1alpha1"
+	"k8s.io/klog"
+)
+
+func (pe *ProbeEvent) deleteBlockDevice(bd blockdevice.BlockDevice, bdAPIList *apis.BlockDeviceList) error {
```
kmova, line 25: Add comment to this function on what it does. Also, probably rename it as either - "handleDeleteEvent" or "deleteBDFromCacheAndDeactivate".

cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -234,166 +231,75 @@ func (pe *ProbeEvent) addBlockDevice(bd blockdevice.BlockDevice) error {
...
-		3. A disk was detached and reconnected to this node
-		4. An add event due to partition table reread . This may cause events to be generated without the disk
-			being physically removed this node - (when a new partition is created on the device also, its the same case)
-	*/
-
-	// check if the disk can be uniquely identified. we try to generate the UUID for the device
-	klog.V(4).Infof("checking if device: %s can be uniquely identified", bd.DevPath)
-	uuid, ok := generateUUID(bd)
-	// if UUID cannot be generated create a GPT partition on the device
+func (pe *ProbeEvent) deleteDeviceFromCache(bd blockdevice.BlockDevice) bool {
```
kmova, line 236: This can also be moved to the deleteDeviceHandler.go

cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -213,14 +197,27 @@ func generateUUID(bd blockdevice.BlockDevice) (string, bool) {
 	return uuid, ok
 }
 
-func (pe *ProbeEvent) addBlockDevice(bd blockdevice.BlockDevice) error {
-
-	bdAPIList, err := pe.Controller.ListBlockDeviceResource(true)
-	if err != nil {
-		klog.Error(err)
-		return err
+// generate old UUID, returns true if the UUID has used path or hostname for generation.
```
kmova, line 200: As we are refactoring this huge file into smaller files, these helper function can be moved to their own file - probably uuidHelpers.go ?

kmova:
@akhilerm -- finally got my slot to review and went through one round. I still need to go through addDeviceHandler.go file. 

My initial impressions on the rest of the files:
- The overall logic seems to work. 
- I really liked the way you are refactoring the eventhandler.go to help with future maintenance. Left a few comments on the naming/refactoring bits. 
- Delete Event Handling is pretty straightforward. We probably need to start maintaining a test lists that we have to cover as we go forwards - for example, handle a delete event for the BD that has several different combinations based on the  states like:
  * active/inactive/unknown
  * legacy-uuid/uuid 
  * claimed/unclaimed/released 
  * with/without partition

cmd/ndm_daemonset/probe/deletehandler.go
```diff
@@ -47,15 +47,22 @@ func (pe *ProbeEvent) deleteBlockDevice(bd blockdevice.BlockDevice, bdAPIList *a
...
+
+	// try to deactivate resource with both UUIDs. This is required in the following case
+
+	existingLegacyBlockDeviceResource := pe.Controller.GetExistingBlockDeviceResource(bdAPIList, bd.UUID)
+	if existingLegacyBlockDeviceResource != nil {
+		pe.Controller.DeactivateBlockDevice(*existingLegacyBlockDeviceResource)
+		klog.V(4).Infof("deactivated device: %s, using legacy UUID", bd.DevPath)
 	}
 
 	existingBlockDeviceResource := pe.Controller.GetExistingBlockDeviceResource(bdAPIList, uuid)
```
pawanpraka1, line 61: we should do this is uuid is valid (ok is true).

cmd/ndm_daemonset/probe/addhandler.go
```diff
@@ -216,3 +241,154 @@ func (pe *ProbeEvent) createBlockDeviceResourceIfNoHolders(bd blockdevice.BlockD
...
+		if legacyBDResource.Annotations == nil {
+			legacyBDResource.Annotations = make(map[string]string)
+		}
+		legacyBDResource.Annotations[internalUUIDAnnotation] = legacyUUIDScheme
+		err := pe.Controller.PushBlockDeviceResource(legacyBDResource, deviceInfo)
+		if err != nil {
+			klog.Errorf("adding %s:%s annotation on %s failed. Error: %v", internalUUIDAnnotation, legacyUUIDScheme, bd.UUID, err)
+			return false, err
+		}
+
```
pawanpraka1, line 285: we should probably return from this if condition?

cmd/ndm_daemonset/probe/addhandler.go
```diff
@@ -216,3 +241,154 @@ func (pe *ProbeEvent) createBlockDeviceResourceIfNoHolders(bd blockdevice.BlockD
...
+		// check if bd with new uuid exists
+		// uses the new algorithm,
+		if pe.Controller.GetExistingBlockDeviceResource(bdAPIList, uuid) != nil {
+			klog.V(4).Infof("device: %s uses GPT based uuid algorithm", bd.DevPath)
+			return true, nil
+		} else {
+			// should never reach this case.
+			// only reaches if manually created blockdevices are present.
+			// this should have been filtered out in filter configs, no firther processing required.
+			return false, nil
```
pawanpraka1, line 308: in case of localpv which is using old blockdevice, which is virtual disk also, generateUUID will return true because of the filesystem. And we will return from here, right?We should proceed further in this case and create the new block device?

cmd/ndm_daemonset/probe/addhandler.go
```diff
@@ -216,3 +241,154 @@ func (pe *ProbeEvent) createBlockDeviceResourceIfNoHolders(bd blockdevice.BlockD
...
+		}
+	}
+
+	// if not in use, there is no need of upgrade.
+	if !bd.DevUse.InUse {
+		return true, nil
+	}
+
+	// device is in use by cstor or localpv
+	klog.V(4).Infof("device: %s in use by cstor / localPV. hence ignoring event", bd.DevPath)
```
pawanpraka1, line 262: ignoring the event?

ndm-operator.yaml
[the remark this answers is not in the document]
pawanpraka1: we should have done this long back. A big thumbs for this change.

kmova:
@akhilerm - with the changes in the PR can you see if the following tests work:

(a) handle the change in the path name
     * setup cstor on /dev/sdb (virtual disk) 
     * rename the /dev/sdb ( as /dev/sdx ) manually in the bd
     * restart ndm 

     Expected Result:
     => NDM should not create a new block device for /dev/sdb 
     => NDM should update the path information to the existing BD

(b) handle the movement of disk from node1 to node2
     * setup cstor on /dev/sdb (virtual disk) (say this is on node-1)
     * move the /dev/sdb (virtual disk) from node 1 to node 2
     * restart ndm on node1 and node2 (in case the changes are not handled automatically)

     Expected Result:
     => NDM should not create a new block device for /dev/sdb 
     => NDM should update the node information to the existing BD

akhilerm:
@kmova @pawanpraka1  [Here](https://gist.github.com/akhilerm/cb9cfff929bff00efaa0d3f70d25a116) is the reference doc for the upgrade flow.

cmd/ndm_daemonset/probe/deleteDeviceHandler.go
akhilerm: No, Consider the following case where the device using legacy UUID has been unclaimed. We will create resource with new UUID. So if it is disconnected we need to deactivate both of them. This is a kind of fail-safe mechanism that is used.

Ideally, after unclaiming blockdevice that used legacy UUID, user should restart the NDM pod , so that everything comes to sync.

cmd/ndm_daemonset/probe/eventhandler.go
akhilerm: Done. Code has been refactored and moved out.

cmd/ndm_daemonset/probe/deletehandler.go
akhilerm: No, for legacy case, the device will have the UUID already generated and stored in the blockDevice UUID field. It is being set by the udev package.

cmd/ndm_daemonset/probe/addhandler.go
akhilerm: Has been modified to only add annotation. New device wont be created. If at the time of upgrade the system is in correct state, i.e the correct BD-BDC relationships are there and used BD is in claimed state. It will only update with FSUUID annotation.

cmd/ndm_daemonset/probe/addhandler.go
akhilerm: refactored / fixed.

akhilerm:
@kmova  Both the above mentioned test cases :
(a) handle the change in the path name and
(b) handle the movement of disk from node1 to node2 are working as expected.

cmd/ndm_daemonset/probe/addhandler.go
```diff
@@ -207,12 +251,344 @@ func (pe *ProbeEvent) createBlockDeviceResourceIfNoHolders(bd blockdevice.BlockD
...
+	err := pe.Controller.CreateBlockDevice(bdAPI)
+	if err != nil {
+		klog.Errorf("unable to push %s (%s) to etcd", bd.UUID, bd.DevPath)
+		return false, err
+	}
+	klog.Infof("Pushed zfs-localPV device: %s (%s) to etcd", bd.UUID, bd.DevPath)
+	return false, nil
+}
+
+// upgradeDeviceInUseByCStor handles the upgrade if the device is used by cstor. returns truw if further processing
```
kmova, line 381: n:_ `true`

cmd/ndm_daemonset/probe/deletehandler.go
```diff
@@ -38,24 +38,63 @@ func (pe *ProbeEvent) removeBlockDeviceFromHierarchyCache(bd blockdevice.BlockDe
...
+		if existingBD != nil {
+			pe.Controller.DeactivateBlockDevice(*existingBD)
+			klog.V(4).Infof("deactivated device: %s, using GPT UUID", bd.DevPath)
+			return nil
+		}
+		// uuid could be generated, but the disk may be using the legacy scheme
+	}
+
+	// try with legacy uuid
+	legacyUUID, _ := generateLegacyUUID(bd)
```
kmova, line 64: This code needs to move towards the end to handle a case like this:
- ndm 0.7 created a BD using legacyuuid for the devices used by zfs local pv
- upgrade to 0.8, a  new BD will be created with partition id
- when the device gets disconnected, only legacyuuid based device will be marked as inactive (noop)
- the new uuid based BD will remain active.

akhilerm: Done.

akhilerm:
@kmova  addressed the review comments and tested the changes.

cmd/ndm_daemonset/probe/deletehandler.go
```diff
@@ -38,24 +38,63 @@ func (pe *ProbeEvent) removeBlockDeviceFromHierarchyCache(bd blockdevice.BlockDe
...
+		pe.Controller.DeactivateBlockDevice(*existingBD)
+		klog.V(4).Infof("deactivated device: %s, using FS UUID annotation", bd.DevPath)
+		return nil
+	}
+
+	// try with partition uuid annotation
+	if existingBD := getExistingBDWithPartitionUUID(bd, bdAPIList); existingBD != nil &&
+		// if the device is a partition, the parent can get deactivated because of search by partition UUID.
+		// Therefore the search result is used only if the device is not a partition.
+		bd.DeviceAttributes.DeviceType != blockdevice.BlockDeviceTypePartition {
```
pawanpraka1, line 84: put this check as a first check in the if condition.

cmd/ndm_daemonset/probe/addhandler.go
```diff
@@ -207,12 +256,344 @@ func (pe *ProbeEvent) createBlockDeviceResourceIfNoHolders(bd blockdevice.BlockD
...
+// is required
+func (pe *ProbeEvent) deviceInUseByMayastor(bd blockdevice.BlockDevice, bdAPIList *apis.BlockDeviceList) (bool, error) {
+	if !bd.DevUse.InUse {
+		return true, nil
+	}
+
+	// not in use by mayastor
+	if bd.DevUse.UsedBy != blockdevice.Mayastor {
+		return true, nil
+	}
```
pawanpraka1, line 332: shouldn't we add BlockDeviceTagLabel to the bd here?

cmd/ndm_daemonset/probe/addhandler.go
```diff
@@ -54,6 +63,36 @@ func (pe *ProbeEvent) addBlockDevice(bd blockdevice.BlockDevice, bdAPIList *apis
...
+	if ok, err := pe.handleUnmanagedDevices(bd, bdAPIList); err != nil {
+		klog.Errorf("error handling unmanaged device %s. error: %v", bd.DevPath, err)
+		return err
+	} else if !ok {
+		klog.V(4).Infof("processed device: %s being used by mayastor/zfs-localPV", bd.DevPath)
+		return nil
+	}
+
+	// if parent device in use, no need to process further
+	if ok, err := pe.isParentDeviceInUse(bd); err != nil {
```
pawanpraka1, line 78: :+1:

cmd/ndm_daemonset/probe/addhandler.go
```diff
@@ -54,6 +63,36 @@ func (pe *ProbeEvent) addBlockDevice(bd blockdevice.BlockDevice, bdAPIList *apis
...
+	// upgrades the devices that are in use and used the legacy method
+	// for uuid generation.
+	if ok, err := pe.upgradeBD(bd, bdAPIList); err != nil {
+		klog.Errorf("upgrade of device: %s failed. Error: %v", bd.DevPath, err)
+		return err
+	} else if !ok {
+		klog.V(4).Infof("device: %s upgraded", bd.DevPath)
+		return nil
+	}
+
```
pawanpraka1, line 95: see if we can break addBlockDevice into smaller part?

cmd/ndm_daemonset/probe/addhandler.go
pawanpraka1: nvm, we are ignoring the creation of BD itself.

cmd/ndm_daemonset/probe/addhandler.go
```diff
@@ -207,12 +256,344 @@ func (pe *ProbeEvent) createBlockDeviceResourceIfNoHolders(bd blockdevice.BlockD
...
+	if ok {
+		existingBD := pe.Controller.GetExistingBlockDeviceResource(bdAPIList, uuid)
+		if existingBD != nil {
+			if existingBD.Status.ClaimState != apis.BlockDeviceUnclaimed {
+				// device in use using gpt UUID
+				return true, nil
+			} else {
+				// should never reach this case
+				klog.Error("unreachable state")
+				return false, nil
```
pawanpraka1, line 455: we should return error here.

cmd/ndm_daemonset/probe/deletehandler.go
akhilerm: Done.

cmd/ndm_daemonset/probe/addhandler.go
akhilerm: will take it up in a later PR @pawanpraka1 .

cmd/ndm_daemonset/probe/addhandler.go
```diff
@@ -207,12 +256,344 @@ func (pe *ProbeEvent) createBlockDeviceResourceIfNoHolders(bd blockdevice.BlockD
...
+			return &bdAPI
+		}
+	}
+	return nil
+}
+
+// getExistingBDWithPartitionUUID returns the blockdevice with matching partition uuid annotation from etcd
+func getExistingBDWithPartitionUUID(bd blockdevice.BlockDevice, bdAPIList *apis.BlockDeviceList) *apis.BlockDevice {
+	for _, bdAPI := range bdAPIList.Items {
+		if len(bd.PartitionInfo.PartitionTableUUID) == 0 {
```
pawanpraka1, line 535: We should return if bd.PartitionInfo.PartitionTableUUID is empty. Do we need to loop through bdAPIList in this case?

akhilerm: Nope. will change it now.

akhilerm: Done. @pawanpraka1 . Both partitionUUID check and FsUUID check has been moved out of the loop.

PR #445, opened 2020-06-30
fix(cleaner); fix cleaning partitions before wiping main disk

akhilerm:

## Pull Request template

When an application creates partitions on a claimed disk, currently only the partition table is wiped as part of
cleanup. There are chances that filesystem can exist on the partitions.  In such a scenario, creating another partition
starting at the same sector as the earlier one, causes system picking up the filesystem signature from the
previous partition and using it.

**What this PR does?**:
To fix the above issue, first the filesystem signature in the partitions need to be wiped and then the partition table. 
This PR modifies the `wipefs` command to list all the partitions of the claimed disk and wipe each one of them, starting with the FS signature in the partitions. Also `partprobe` command is invoked so as to re-read the partition table on the disk after wiping.

No

Yes. Created a cstor pool on a raw disk. Then deleted the pool. After that partition was manually created. System is not reporting
any known filesystems, earlier `zfs_member` filesystem was reported.

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/pkg/cleaner/jobcontroller.go
+++ b/pkg/cleaner/jobcontroller.go
@@ -22,6 +22,7 @@ package cleaner
 
 import (
 	"context"
+	"fmt"
 	"github.com/openebs/node-disk-manager/blockdevice"
 	"github.com/openebs/node-disk-manager/cmd/ndm_daemonset/controller"
 	"github.com/openebs/node-disk-manager/pkg/apis/openebs/v1alpha1"
@@ -39,8 +40,6 @@ const (
 	JobNamePrefix = "cleanup-"
 	// BDLabel is the label set on the job for identification of the BD
 	BDLabel = "blockdevice"
-	// BlockCleanerCommand is the command used to clean raw block
-	BlockCleanerCommand = "wipefs"
 )
 
 // JobController defines the interface for the job controller.
@@ -75,10 +74,19 @@ func NewCleanupJob(bd *v1alpha1.BlockDevice, volMode VolumeMode, tolerations []v
 	mountName := "vol-mount"
 
 	if volMode == VolumeModeBlock {
+		jobContainer.Command = []string{"/bin/sh", "-c"}
+
+		// fdisk is used to get all the partitions of the device.
+		// first all the partitions are cleared off any filesystem signatures, then the actual partition table
+		// header is removed. partprobe is called so as to re-read partition table, and update system with
+		// the changes/
 		// wipefs erases the filesystem signature from the block
 		// -a    wipe all magic strings
 		// -f    force erasure
-		jobContainer.Command = getCommand(BlockCleanerCommand, "-af", bd.Spec.Path)
+		args := fmt.Sprintf("(fdisk -o Device -l %s | grep \"^%s\" | sort -r | xargs -I '{}' wipefs -fa '{}') && wipefs -fa %s && partprobe %s",
+			bd.Spec.Path, bd.Spec.Path, bd.Spec.Path, bd.Spec.Path)
+
+		jobContainer.Args = []string{args}
 
 		// in case of sparse disk, need to mount the sparse file directory
 		// and clear the sparse file
@@ -197,15 +205,6 @@ func (c *jobController) CancelJob(bdName string) error {
 
 func generateCleaningJobName(bdName string) string {
 	return JobNamePrefix + bdName
-}
-
-func getCommand(cmd string, args ...string) []string {
-	var command []string
-	command = append(command, cmd)
-	for _, arg := range args {
-		command = append(command, arg)
-	}
-	return command
 }
 
 // GetNodeName gets the Node name from BlockDevice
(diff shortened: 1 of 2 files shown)


pkg/cleaner/jobcontroller.go
```diff
@@ -75,10 +74,20 @@ func NewCleanupJob(bd *v1alpha1.BlockDevice, volMode VolumeMode, tolerations []v
...
+		// if deviceType is a disk, there can be partitions on the disk. In that case
+		// partitions need to be wiped before wiping main disk
+		if bd.Spec.Details.DeviceType == "disk" {
+			pathRegex = bd.Spec.Path + "*"
+		}
 		// wipefs erases the filesystem signature from the block
 		// -a    wipe all magic strings
 		// -f    force erasure
-		jobContainer.Command = getCommand(BlockCleanerCommand, "-af", bd.Spec.Path)
+		args := fmt.Sprintf("ls -r %s | xargs wipefs -af", pathRegex)
```
kmova, line 88: Can the ls command be made more specific for partitions? For example, what if we have the case like: `/dev/sda` and `/dev/sdaa` .  A case where mounted volumes (like iSCSI) start getting a new device mount.

pkg/cleaner/jobcontroller.go
```diff
@@ -75,10 +74,20 @@ func NewCleanupJob(bd *v1alpha1.BlockDevice, volMode VolumeMode, tolerations []v
 	mountName := "vol-mount"
 
 	if volMode == VolumeModeBlock {
+		jobContainer.Command = []string{"/bin/sh", "-c"}
+
+		pathRegex := bd.Spec.Path
+		// if deviceType is a disk, there can be partitions on the disk. In that case
+		// partitions need to be wiped before wiping main disk
+		if bd.Spec.Details.DeviceType == "disk" {
+			pathRegex = bd.Spec.Path + "*"
```
pawanpraka1, line 83: can we traverse the children and wipe the device path here?

pkg/cleaner/jobcontroller.go
akhilerm: Will fix it @kmova

pkg/cleaner/jobcontroller.go
akhilerm: You mean, for `sdb`, traversing through `sdb1`, `sdb2`... etc and wiping them right?

pawanpraka1: Yes, instead of doing path*, we should traverse the children and wipe it.

pkg/cleaner/jobcontroller.go
akhilerm: Done.

pkg/cleaner/jobcontroller.go
akhilerm: Done. Thank you.

pkg/cleaner/jobcontroller.go
```diff
@@ -75,10 +74,19 @@ func NewCleanupJob(bd *v1alpha1.BlockDevice, volMode VolumeMode, tolerations []v
...
+
+		// fdisk is used to get all the partitions of the device.
+		// first all the partitions are cleared off any filesystem signatures, then the actual partition table
+		// header is removed. partprobe is called so as to re-read partition table, and update system with
+		// the changes/
 		// wipefs erases the filesystem signature from the block
 		// -a    wipe all magic strings
 		// -f    force erasure
-		jobContainer.Command = getCommand(BlockCleanerCommand, "-af", bd.Spec.Path)
+		args := fmt.Sprintf("(fdisk -o Device -l %s | grep \"^%s\" | sort -r | xargs -I '{}' wipefs -fa '{}') && wipefs -fa %s && partprobe %s",
```
pawanpraka1, line 86: we can remove sort  as it is not needed here.

pkg/cleaner/jobcontroller.go
```diff
@@ -75,10 +74,19 @@ func NewCleanupJob(bd *v1alpha1.BlockDevice, volMode VolumeMode, tolerations []v
 	mountName := "vol-mount"
 
 	if volMode == VolumeModeBlock {
+		jobContainer.Command = []string{"/bin/sh", "-c"}
+
+		// fdisk is used to get all the partitions of the device.
+		// first all the partitions are cleared off any filesystem signatures, then the actual partition table
+		// header is removed. partprobe is called so as to re-read partition table, and update system with
```
pawanpraka1, line 81: so partprobe is working?

akhilerm: yes. if we give `partprobe /dev/sdX` its working. Earlier I tried using just `partprobe` command

pkg/cleaner/jobcontroller.go
akhilerm: Done.

akhilerm:
> looks good. test the command for other type of disks also (like nvme and any other we have).

Already tested on nvme drive. It works as expected.

pawanpraka1:
> Already tested on nvme drive. It works as expected.

Awesome!!

PR #430, opened 2020-06-01
feat(probe): add a new probe to check if device is in use

akhilerm:
**Why is this PR required? What issue does it fix?**:
Currently LocalPV and cstor are the 2 storage engines that use NDM. But other engines like ZFS-localPV and mayastor can also co-exist in the same node. This leads to NDM assiging already used devices to cstor or localPV. Therefore, some mechanism is required to read the disk and check if it is used by any storage engines

**What this PR does?**:
This PR adds a new probe which uses mount path and blkid to find the storage engine that is using a disk.

**Does this PR require any upgrade changes?**:
No

**If the changes in this PR are manually verified, list down the scenarios covered:**:

**Any additional information for your reviewer?** : 
_Mention if this PR is part of any design or a continuation of previous PRs_

**Checklist:**
- [x] Fixes [#3042](https://github.com/openebs/openebs/issues/3042)
- [x] PR Title follows the convention of  `<type>(<scope>): <subject>`
- [x] Has the change log section been updated? 
- [ ] Commit has unit tests
- [ ] Commit has integration tests
- [ ] (Optional) Are upgrade changes included in this PR? If not, mention the issue/PR to track: 
- [ ] (Optional) If documentation changes are required, which issue on https://github.com/openebs/openebs-docs is used to track them:

PR #450, opened 2020-07-06
fix(probe): close open fd after exclusive open

akhilerm:

NDM was keeping a file descriptor open after an exclusive open, causing other applications to receive a resource busy error.

**What this PR does?**:
Closes the open FD
fixes https://github.com/openebs/openebs/issues/3079

No

Bug was introduced in used-by probe as part of https://github.com/openebs/node-disk-manager/pull/430

**Checklist:**
- [x] Fixes [#3079](https://github.com/openebs/openebs/issues/3079)
- [ ] Commit has unit tests

### DIFF
--- a/cmd/ndm_daemonset/probe/usedbyprobe.go
+++ b/cmd/ndm_daemonset/probe/usedbyprobe.go
@@ -200,7 +200,7 @@ func getBlockDeviceZFSPartition(bd blockdevice.BlockDevice) (string, bool) {
 // isBlockDeviceInUseByKernel tries to open the device exclusively to check if the device is
 // being held by some process. eg: If kernel zfs uses the disk, the open will fail
 func isBlockDeviceInUseByKernel(path string) (bool, error) {
-	_, err := os.OpenFile(path, os.O_EXCL, 0444)
+	f, err := os.OpenFile(path, os.O_EXCL, 0444)
 
 	if errors.Is(err, syscall.EBUSY) {
 		return true, nil
@@ -208,5 +208,6 @@ func isBlockDeviceInUseByKernel(path string) (bool, error) {
 	if err != nil {
 		return false, err
 	}
+	defer f.Close()
 	return false, nil
 }
--- a/changelogs/unreleased/450-akhilerm
+++ b/changelogs/unreleased/450-akhilerm
@@ -1,0 +1,1 @@
+fix bug of having an open file descriptor in NDM causing applications to receive resource busy error.


cmd/ndm_daemonset/probe/usedbyprobe.go
```diff
@@ -200,13 +200,14 @@ func getBlockDeviceZFSPartition(bd blockdevice.BlockDevice) (string, bool) {
 // isBlockDeviceInUseByKernel tries to open the device exclusively to check if the device is
 // being held by some process. eg: If kernel zfs uses the disk, the open will fail
 func isBlockDeviceInUseByKernel(path string) (bool, error) {
-	_, err := os.OpenFile(path, os.O_EXCL, 0444)
+	f, err := os.OpenFile(path, os.O_EXCL, 0444)
 
 	if errors.Is(err, syscall.EBUSY) {
```
vishnuitta, line 205: this just means that its in use by someone, but, need not be by kernel

akhilerm: @vishnuitta Whats the best way to check if disk is in use by kernel zfs?

cmd/ndm_daemonset/probe/usedbyprobe.go
```diff
@@ -200,13 +200,14 @@ func getBlockDeviceZFSPartition(bd blockdevice.BlockDevice) (string, bool) {
...
 		return true, nil
 	}
 	if err != nil {
 		return false, err
 	}
+	defer f.Close()
```
kmova, line 211: should there be a nil check for `f` ?

akhilerm: `f.Close()` [internally](https://golang.org/src/os/file_unix.go?s=7277:7305#L223) has a nil check.

cmd/ndm_daemonset/probe/usedbyprobe.go
vishnuitta: how about running zpool commands to find the disks in use by kernel zfs? its also good to change name of the function

cmd/ndm_daemonset/probe/usedbyprobe.go
mittachaitu, line 211: Since we are not doing any activity after opening the file(ex: re-opening the file with the same flags which leads to point to the same fd) it is assumed the `Close` func will not return any error.

cmd/ndm_daemonset/probe/usedbyprobe.go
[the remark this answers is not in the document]
akhilerm: Caller is logging the error @mittachaitu

cmd/ndm_daemonset/probe/usedbyprobe.go
akhilerm: yes @mittachaitu . We are just opening the file and if open succeeds, we close the file.

PR #455, opened 2020-07-10
refactor(code): Exported rescan functionality

harshthakur9030:

## Pull Request template

This PR exposes the method which is used to sync etcd and ndm local state

**What this PR does?**:
Exposes rescan method which helps to sync etcd and ndm local state. It also ensures that the udev method being called is thread safe. 

No

Tested it on a GKE cluster.  

This is required for the gRPC PR.

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/changelogs/unreleased/455-harshthakur9030
+++ b/changelogs/unreleased/455-harshthakur9030
@@ -1,0 +1,1 @@
+Make udev scan operation thread safe.
--- a/cmd/ndm_daemonset/probe/eventhandler.go
+++ b/cmd/ndm_daemonset/probe/eventhandler.go
@@ -18,6 +18,7 @@ package probe
 
 import (
 	"fmt"
+
 	"github.com/openebs/node-disk-manager/blockdevice"
 	"github.com/openebs/node-disk-manager/cmd/ndm_daemonset/controller"
 	apis "github.com/openebs/node-disk-manager/pkg/apis/openebs/v1alpha1"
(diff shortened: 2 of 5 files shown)


cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -141,19 +142,20 @@ func (pe *ProbeEvent) deleteBlockDeviceEvent(msg controller.EventMessage) {
...
-		go pe.initOrErrorEvent()
+		go pe.Rescan()
 	}
 }
 
-// initOrErrorEvent rescan system and update disk resource this is
-// used for initial setup and when any uid mismatch or error occurred.
-func (pe *ProbeEvent) initOrErrorEvent() {
+// Rescan syncs etcd and NDM
+func (pe *ProbeEvent) Rescan() error {
```
akhilerm, line 150: @harshthakur9030  Does it make sense to move Rescan out of this and keep it completely in the udevprobe. That way you tie the rescan with udev,

harshthakur9030: I think it's better if we keep it in event handler itself as a probe event method as probe event struct is simpler to initialize than a udev probe struct(constructor for which isn't exported) so considering Rescan is actually just a wrapper around the udev scan, it's fine I suppose.

akhilerm: `udev.Rescan(c Controller)`, something like this, because we have the controller. Initialising both Probevent and rescan only requires the pointer to controller.

harshthakur9030: Makes sense. Will refactor this then

changelogs/unreleased/455-harshthakur9030
```diff
@@ -0,0 +1 @@
+Exported rescan functionality
```
akhilerm, line 1: will be a better wording, if you can use.
```suggestion
Make udev scan operation thread safe.
```

PR #461, opened 2020-07-16
feat(metrics): exporting seachest data to prometheus

### LINKED ISSUE #451: Exporter : Expose exisiting seachest data to prometheus
**Describe the problem/challenge you have**
Currently, there's no seachest data except temperature being exposed by prometheus.

**Describe the solution you'd like**
Expose more seachest data like latency, R/W bytes, etc

**Anything else you would like to add:**
Following data is already available :
* HighestTemperature
* LowestTempearature
* RotationalLatency
* Capacity
* LogicalSectorSize
* PhysicalSectorSize
* RotationRate
* DriveType
* TotalBytesRead
* TotalBytesWritten
* DeviceUtilization

**Environment:**
All.

harshthakur9030:

## Pull Request template

Exports seachest data which is already fetched

**What this PR does?**:
Exports seachest data to prometheus

No

The following data is exported :
    HighestTemperature
    LowestTempearature
    RotationalLatency
    Capacity
    LogicalSectorSize
    PhysicalSectorSize
    RotationRate
    DriveType
    TotalBytesRead
    TotalBytesWritten
    DeviceUtilization
Other than the above, Rotational Latency is calculated and exported

**Checklist:**
- [x] Fixes #451 
- [ ] Commit has unit tests

### DIFF
--- a/blockdevice/blockdevice.go
+++ b/blockdevice/blockdevice.go
@@ -119,11 +119,35 @@ type BlockDevice struct {
 	// DependentDevices stores the dependent devices
 	DependentDevices DependentBlockDevices
 
-	// TemperatureInfo stores the temperature information of the drive
-	TemperatureInfo TemperatureInformation
+	SMARTInfo SMARTStats
 
 	// Status contains the state of the blockdevice
 	Status Status
+}
+
+// SMARTStats represents stats from SMART spec and data fetched/calculated by data from seachest
+type SMARTStats struct {
+
+	// RotationRate stores the rotation rate of a block device, only applicable for HDD
+	RotationRate uint16
+
+	// RotationalLatency stores the latency of the drive, only applicable for HDD
+	RotationalLatency float64
+
+	// TemperatureInfo stores the temperature information of the drive
+	TemperatureInfo TemperatureInformation
+
+	// TotalBytesRead stores the total amount of bytes read by a block device
+	TotalBytesRead uint64
+
+	// TotalBytesWritten stores the total amount of bytes written by a block device
+	TotalBytesWritten uint64
+
+	// UtilizationRate stores the rate of utilization of a block device
+	UtilizationRate float64
+
+	// PercentEnduranceUsed stores the endurance used in percent
+	PercentEnduranceUsed float64
 }
 
 // Identifier represents the various identifiers that can be used to
@@ -276,12 +300,26 @@ type DevLink struct {
 
 // TemperatureInformation stores the temperature information of the blockdevice
 type TemperatureInformation struct {
-	// TemperatureDataValid specifies whether the current temperature
+	// CurrentTemperatureDataValid specifies whether the current temperature
 	// data reported is valid or not
-	TemperatureDataValid bool
+	CurrentTemperatureDataValid bool
+
+	// LowestTemperatureDataValid specifies whether the lowest temperature
+	// data reported is valid or not
+	LowestTemperatureDataValid bool
+
+	// HighestTemperatureDataValid specifies whether the highest temperature
+	// data reported is valid or not
+	HighestTemperatureDataValid bool
 
 	// CurrentTemperature is the temperature of the drive in celsius
 	CurrentTemperature int16
+
+	// LowestTemperature is the lowest temperature of the drive recorded in celsius
+	LowestTemperature int16
+
+	// HighestTemperature is the highest temperature of the drive recorded in celsius
+	HighestTemperature int16
 }
 
 // PartitionInformation contains information related to the partition, if this
--- a/cmd/ndm_daemonset/probe/seachestprobe.go
+++ b/cmd/ndm_daemonset/probe/seachestprobe.go
@@ -151,6 +151,7 @@ func (scp *seachestProbe) FillBlockDeviceDetails(blockDevice *blockdevice.BlockD
 
 	blockDevice.SMARTInfo.TemperatureInfo.CurrentTemperatureDataValid = seachestProbe.
 		SeachestIdentifier.GetTemperatureDataValidStatus(driveInfo)
+
 	klog.V(4).Infof("Disk: %s TemperatureDataValid:%t filled by seachest.",
 		blockDevice.DevPath, blockDevice.SMARTInfo.TemperatureInfo.CurrentTemperatureDataValid)
 
@@ -161,24 +162,29 @@ func (scp *seachestProbe) FillBlockDeviceDetails(blockDevice *blockdevice.BlockD
 		klog.V(4).Infof("Disk: %s CurrentTemperature:%d filled by seachest.",
 			blockDevice.DevPath, blockDevice.SMARTInfo.TemperatureInfo.CurrentTemperature)
 
-		blockDevice.SMARTInfo.TemperatureInfo.HighestTemperatureDataValid = seachestProbe.
-			SeachestIdentifier.GetHighestValid(driveInfo)
+	}
 
-		klog.V(4).Infof("Disk: %s HighestTemperatureDataValid:%t filled by seachest.",
-			blockDevice.DevPath, blockDevice.SMARTInfo.TemperatureInfo.HighestTemperatureDataValid)
+	blockDevice.SMARTInfo.TemperatureInfo.HighestTemperatureDataValid = seachestProbe.
+		SeachestIdentifier.GetHighestValid(driveInfo)
+
+	klog.V(4).Infof("Disk: %s HighestTemperatureDataValid:%t filled by seachest.",
+		blockDevice.DevPath, blockDevice.SMARTInfo.TemperatureInfo.HighestTemperatureDataValid)
+
+	if blockDevice.SMARTInfo.TemperatureInfo.HighestTemperatureDataValid == true {
 
 		blockDevice.SMARTInfo.TemperatureInfo.HighestTemperature = seachestProbe.
 			SeachestIdentifier.GetHighestTemperature(driveInfo)
 
 		klog.V(4).Infof("Disk: %s HighestTemperature:%d filled by seachest.",
 			blockDevice.DevPath, blockDevice.SMARTInfo.TemperatureInfo.HighestTemperature)
+	}
+	blockDevice.SMARTInfo.TemperatureInfo.LowestTemperatureDataValid = seachestProbe.
+		SeachestIdentifier.GetLowestValid(driveInfo)
 
-		blockDevice.SMARTInfo.TemperatureInfo.LowestTemperatureDataValid = seachestProbe.
-			SeachestIdentifier.GetLowestValid(driveInfo)
+	klog.V(4).Infof("Disk: %s LowestValid:%t filled by seachest.",
+		blockDevice.DevPath, blockDevice.SMARTInfo.TemperatureInfo.LowestTemperatureDataValid)
 
-		klog.V(4).Infof("Disk: %s LowestValid:%t filled by seachest.",
-			blockDevice.DevPath, blockDevice.SMARTInfo.TemperatureInfo.LowestTemperatureDataValid)
-
+	if blockDevice.SMARTInfo.TemperatureInfo.LowestTemperatureDataValid == true {
 		blockDevice.SMARTInfo.TemperatureInfo.LowestTemperature = seachestProbe.
 			SeachestIdentifier.GetLowestTemperature(driveInfo)
 
--- a/cmd/ndm_daemonset/probe/seachestprobe.go
+++ b/cmd/ndm_daemonset/probe/seachestprobe.go
@@ -124,65 +124,65 @@ func (scp *seachestProbe) FillBlockDeviceDetails(blockDevice *blockdevice.BlockD
 
 	// All the below mentioned fields will be filled in only after BlockDevice struct
 	// starts supporting them.
-	/*if d.RotationRate == 0 {
-		d.RotationRate = seachestProbe.SeachestIdentifier.GetRotationRate(driveInfo)
-		klog.V(4).Infof("Disk: %s RotationRate:%d filled by seachest.", d.Path, d.RotationRate)
-	}*/
-
-	/*if d.TotalBytesRead == 0 {
-		d.TotalBytesRead = seachestProbe.SeachestIdentifier.GetTotalBytesRead(driveInfo)
-		klog.V(4).Infof("Disk: %s TotalBytesRead:%d filled by seachest.", d.Path, d.TotalBytesRead)
+	if blockDevice.SMARTInfo.RotationRate == 0 {
+		blockDevice.SMARTInfo.RotationRate = seachestProbe.SeachestIdentifier.GetRotationRate(driveInfo)
+		klog.V(4).Infof("Disk: %s RotationRate:%d filled by seachest.", blockDevice.DevPath, blockDevice.SMARTInfo.RotationRate)
 	}
 
-	if d.TotalBytesWritten == 0 {
-		d.TotalBytesWritten = seachestProbe.SeachestIdentifier.GetTotalBytesWritten(driveInfo)
-		klog.V(4).Infof("Disk: %s TotalBytesWritten:%d filled by seachest.", d.Path, d.TotalBytesWritten)
+	if blockDevice.SMARTInfo.TotalBytesRead == 0 {
+		blockDevice.SMARTInfo.TotalBytesRead = seachestProbe.SeachestIdentifier.GetTotalBytesRead(driveInfo)
+		klog.V(4).Infof("Disk: %s TotalBytesRead:%d filled by seachest.", blockDevice.DevPath, blockDevice.SMARTInfo.TotalBytesRead)
 	}
 
-	if d.DeviceUtilizationRate == 0 {
-		d.DeviceUtilizationRate = seachestProbe.SeachestIdentifier.GetDeviceUtilizationRate(driveInfo)
-		klog.V(4).Infof("Disk: %s DeviceUtilizationRate:%f filled by seachest.", d.Path, d.DeviceUtilizationRate)
+	if blockDevice.SMARTInfo.TotalBytesWritten == 0 {
+		blockDevice.SMARTInfo.TotalBytesWritten = seachestProbe.SeachestIdentifier.GetTotalBytesWritten(driveInfo)
+		klog.V(4).Infof("Disk: %s TotalBytesWritten:%d filled by seachest.", blockDevice.DevPath, blockDevice.SMARTInfo.TotalBytesWritten)
 	}
 
-	if d.PercentEnduranceUsed == 0 {
-		d.PercentEnduranceUsed = seachestProbe.SeachestIdentifier.GetPercentEnduranceUsed(driveInfo)
-		klog.V(4).Infof("Disk: %s PercentEnduranceUsed:%f filled by seachest.", d.Path, d.PercentEnduranceUsed)
-	}*/
+	if blockDevice.SMARTInfo.UtilizationRate == 0 {
+		blockDevice.SMARTInfo.UtilizationRate = seachestProbe.SeachestIdentifier.GetDeviceUtilizationRate(driveInfo)
+		klog.V(4).Infof("Disk: %s UtilizationRate:%f filled by seachest.", blockDevice.DevPath, blockDevice.SMARTInfo.UtilizationRate)
+	}
 
-	blockDevice.TemperatureInfo.TemperatureDataValid = seachestProbe.
+	if blockDevice.SMARTInfo.PercentEnduranceUsed == 0 {
+		blockDevice.SMARTInfo.PercentEnduranceUsed = seachestProbe.SeachestIdentifier.GetPercentEnduranceUsed(driveInfo)
+		klog.V(4).Infof("Disk: %s PercentEnduranceUsed:%f filled by seachest.", blockDevice.DevPath, blockDevice.SMARTInfo.PercentEnduranceUsed)
+	}
+
+	blockDevice.SMARTInfo.TemperatureInfo.CurrentTemperatureDataValid = seachestProbe.
 		SeachestIdentifier.GetTemperatureDataValidStatus(driveInfo)
 	klog.V(4).Infof("Disk: %s TemperatureDataValid:%t filled by seachest.",
-		blockDevice.DevPath, blockDevice.TemperatureInfo.TemperatureDataValid)
+		blockDevice.DevPath, blockDevice.SMARTInfo.TemperatureInfo.CurrentTemperatureDataValid)
 
-	if blockDevice.TemperatureInfo.TemperatureDataValid == true {
-		blockDevice.TemperatureInfo.CurrentTemperature = seachestProbe.
+	if blockDevice.SMARTInfo.TemperatureInfo.CurrentTemperatureDataValid == true {
+		blockDevice.SMARTInfo.TemperatureInfo.CurrentTemperature = seachestProbe.
 			SeachestIdentifier.GetCurrentTemperature(driveInfo)
 
 		klog.V(4).Infof("Disk: %s CurrentTemperature:%d filled by seachest.",
-			blockDevice.DevPath, blockDevice.TemperatureInfo.CurrentTemperature)
+			blockDevice.DevPath, blockDevice.SMARTInfo.TemperatureInfo.CurrentTemperature)
 
-		/*d.TemperatureInfo.HighestValid = seachestProbe.
+		blockDevice.SMARTInfo.TemperatureInfo.HighestTemperatureDataValid = seachestProbe.
 			SeachestIdentifier.GetHighestValid(driveInfo)
 
-		klog.V(4).Infof("Disk: %s HighestValid:%t filled by seachest.",
-			d.Path, d.TemperatureInfo.HighestValid)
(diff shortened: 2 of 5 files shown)


pkg/metrics/smart/metrics.go
```diff
@@ -64,11 +103,26 @@ func (m *Metrics) Collectors() []prometheus.Collector {
...
 		m.blockDeviceCurrentTemperature,
+		m.blockDeviceRotationalLatency,
+		m.blockDeviceRotationRate,
+		m.blockDeviceHighestTemperature,
+		m.blockDeviceLowestTemperature,
+		m.blockDeviceLogicalSectorSize,
+		m.blockDevicePhysicalSectorSize,
+		m.blockDeviceTotalReadBytes,
+		m.blockDeviceTotalWrittenBytes,
+		m.blockDeviceTotalReadBytes,
```
utkarshmani1997, line 114: Duplicate blockDeviceTotalReadBytes

pkg/metrics/smart/metrics.go
```diff
@@ -64,11 +103,26 @@ func (m *Metrics) Collectors() []prometheus.Collector {
...
+		m.blockDeviceTotalWrittenBytes,
```
utkarshmani1997, line 115: Duplicate blockDeviceTotalWrittenBytes

blockdevice/blockdevice.go
```diff
@@ -253,6 +277,9 @@ type DeviceAttribute struct {
 	// Eg : PersistentDisk, Virtaul_disk, QEMU_HARDDISK, EphemeralDisk
 	Model string
 
+	//Model number of the device
+	ModelNumber string
```
akhilerm, line 281: question: What is the difference between model and model number?

blockdevice/blockdevice.go
```diff
@@ -282,6 +309,12 @@ type TemperatureInformation struct {
 
 	// CurrentTemperature is the temperature of the drive in celsius
 	CurrentTemperature int16
+
+	// LowestTemperature is the temperature of the drive in celsius
+	LowestTemperature int16
+
+	// HighestTemperature is the temperature of the drive in celsius
```
akhilerm, line 316: Comments should describe what the fields are and what information they convey. Since this is the only place where all blockdevice information is stored, the comments can be pretty descriptive of what the field is.

ndm-exporter/collector/seachest.go
```diff
@@ -52,8 +52,23 @@ type SeachestCollector struct {
...
 type SeachestMetricData struct {
-	SeachestIdentifier *seachest.Identifier
-	TempInfo           blockdevice.TemperatureInformation
+	SeachestIdentifier   *seachest.Identifier
+	TempInfo             blockdevice.TemperatureInformation
+	ModelNumber          string
+	SerialNumber         string
+	VendorID             string
+	FirmwareVersion      string
+	UUID                 string
```
akhilerm, line 61: I think UUID reported by seachest is the WWN of the disk. In that case, the field should be renamed.

ndm-exporter/collector/seachest.go
```diff
@@ -52,8 +52,23 @@ type SeachestCollector struct {
 // SeachestMetricData is the struct which holds the data from seachest library
 // corresponding to each blockdevice
 type SeachestMetricData struct {
-	SeachestIdentifier *seachest.Identifier
-	TempInfo           blockdevice.TemperatureInformation
+	SeachestIdentifier   *seachest.Identifier
+	TempInfo             blockdevice.TemperatureInformation
+	ModelNumber          string
```
akhilerm, line 57: Model or ModelNumber?

ndm-exporter/collector/seachest.go
```diff
@@ -52,8 +52,23 @@ type SeachestCollector struct {
...
+	SerialNumber         string
+	VendorID             string
+	FirmwareVersion      string
+	UUID                 string
+	Capacity             uint64
+	LogicalSectorSize    uint32
+	PhysicalSectorSize   uint32
+	RotationRate         uint16
+	Latency              float64
```
akhilerm, line 66: q: Is this same as rotational latency in the metrics?

pkg/metrics/smart/metrics.go
```diff
@@ -31,17 +31,56 @@ type MetricsData struct {
...
+	UUID            string
+	Path            string
+	HostName        string
+	NodeName        string
+	ModelNumber     string
+	Model           string
+	SerialNumber    string
+	VendorID        string
+	FirmwareVersion string
+	DriveType       string
```
akhilerm, line 83: Why are the labels changed? The older labels were selected so as to connect/merge metrics from node exporter and NDM exporter. 
i.e Those labels help to map a `blockdevice-xxx` resource to corresponding metrics from prometheus.

pkg/seachest/seachest.go
```diff
@@ -165,6 +166,18 @@ func (I *Identifier) GetRotationRate(driveInfo *C.driveInformationSAS_SATA) uint
 	return 0
 }
 
+func (I *Identifier) GetRotationalLatency(driveInfo *C.driveInformationSAS_SATA) float64 {
+	if driveInfo.rotationRate > 1 {
+		// latency = 1/rpm/60
+		klog.Infof("Device is : %v", I.DevPath)
+		rpm := (uint16)(driveInfo.rotationRate)
```
akhilerm, line 173: Provide a reference to this, so that we can see why the formula was chosen

ndm-exporter/collector/seachest.go
```diff
@@ -52,8 +52,23 @@ type SeachestCollector struct {
...
+	SeachestIdentifier   *seachest.Identifier
+	TempInfo             blockdevice.TemperatureInformation
+	ModelNumber          string
+	SerialNumber         string
+	VendorID             string
+	FirmwareVersion      string
+	UUID                 string
+	Capacity             uint64
+	LogicalSectorSize    uint32
+	PhysicalSectorSize   uint32
```
akhilerm, line 64: These fields are fixed for a drive and will not change. Also this information is available in the etcd. For pulling metrics from etcd we have the ndm-cluster-exporter and the static collector. 

These fields, and any other field that does not change in values / value is available in etcd should be fetched via the static collector. This reduces pressure on the disk, since each metrics collection involves some SCSI commands to the disk.

blockdevice/blockdevice.go
```diff
@@ -119,13 +119,37 @@ type BlockDevice struct {
...
+
+	// TotalBytesWritten stores the total amount of bytes read by a block device
+	TotalBytesWritten uint64
+
+	// UtilizationRate stores the rate of utilization of a block device
+	UtilizationRate float64
+
+	// PercentEnduranceUsed stores the endurance used in percent
+	PercentEnduranceUsed float64
+}
```
akhilerm, line 151: :+1:

ndm-exporter/collector/seachest.go
```diff
@@ -66,8 +73,17 @@ func NewSeachestMetricCollector(c kubernetes.Client) prometheus.Collector {
 	}
 	sc.metrics.WithBlockDeviceCurrentTemperature().
 		WithBlockDeviceCurrentTemperatureValid().
+		WithBlockDeviceHighestTemperature().
+		WithBlockDeviceLowestTempearature().
 		WithRejectRequest().
-		WithErrorRequest()
+		WithErrorRequest().
```
akhilerm, line 79: nit: keep error and reject request as the last items.

ndm-exporter/collector/seachest.go
```diff
@@ -185,8 +211,30 @@ func (sc *SeachestMetricData) getSeachestData() error {
 		klog.Errorf("error fetching basic disk info using seachest. %s", seachest.SeachestErrors(err))
 		return fmt.Errorf("error getting seachest data for metrics. %s", seachest.SeachestErrors(err))
 	}
+
 	sc.TempInfo.TemperatureDataValid = sc.SeachestIdentifier.GetTemperatureDataValidStatus(driveInfo)
 	sc.TempInfo.CurrentTemperature = sc.SeachestIdentifier.GetCurrentTemperature(driveInfo)
+	sc.TempInfo.LowestTemperature = sc.SeachestIdentifier.GetLowestTemperature(driveInfo)
+	sc.TempInfo.HighestTemperature = sc.SeachestIdentifier.GetHighestTemperature(driveInfo)
```
akhilerm, line 218: The temperature fetching code is commented out in seachest probe. May need to uncomment it.

pkg/metrics/smart/metrics.go
```diff
@@ -31,6 +31,33 @@ type MetricsData struct {
...
+	blockDeviceLowestTemperature *prometheus.GaugeVec
+
+	// blockDeviceRotationalLatency is rotational latency of a block device
+	blockDeviceRotationalLatency *prometheus.GaugeVec
+
+	// blockDeviceCapacity is capacity of block device
+	blockDeviceCapacity *prometheus.GaugeVec
+
+	// blockDeviceRotationRate is rotation rate of a block device in RPM
+	blockDeviceRotationRate *prometheus.GaugeVec
```
akhilerm, line 47: The rotation rate is a fixed value. Should we collect it from here? Or use the value available in etcd?

pkg/metrics/smart/metrics.go
```diff
@@ -64,11 +91,21 @@ func (m *Metrics) Collectors() []prometheus.Collector {
...
+		m.blockDeviceTotalReadBytes,
+		m.blockDeviceTotalWrittenBytes,
+		m.blockDeviceUtilizationRate,
+		m.blockDevicePercentEnduranceUsed,
 		m.rejectRequestCount,
 		m.errorRequestCount,
 	}
 }
 
+var labels []string = []string{"blockdevicename", "path", "hostname", "nodename"}
```
akhilerm, line 107: :+1:

pkg/seachest/seachest.go
akhilerm: @harshthakur9030  Is this going to be the formula we are going to use?. Because still we dont have SSDs handled here.

akhilerm (commented):
provided some review comments.

pkg/metrics/smart/metrics.go
harshthakur9030: I think it's better if it's here as cloud providers' virtual disks show a change in performance when there's a change in capacity.

pkg/seachest/seachest.go
harshthakur9030: SSD won't have a rotation rate. If it's 0, users can understand that it's an SSD.

ndm-exporter/collector/seachest.go
harshthakur9030: @akhilerm Just checked, it's not commented.

pkg/metrics/smart/metrics.go
akhilerm: The rotation rate is something like 5400 RPM, 7200 RPM right? Even if capacity changes that value won;t change.

pkg/metrics/smart/metrics.go
```diff
@@ -169,18 +326,42 @@ func (m *Metrics) SetBlockDeviceCurrentTemperature(currentTemp int16) *Metrics {
...
 }
 
+// SetBlockDeviceHighestTemperature sets the highest temperature value to the metric
+func (m *Metrics) SetBlockDeviceHighestTemperature(highTemp int16) *Metrics {
+	m.blockDeviceHighestTemperature.WithLabelValues(m.UUID,
+		m.Path,
+		m.HostName,
+		m.NodeName,
+	).
+		Set(float64(highTemp))
```
akhilerm, line 342: Can you check for `gofmt` errors here. I think the `Set(` method is misaligned.

akhilerm: The same comment applies for all the `Set()` methods.

pkg/seachest/seachest.go
akhilerm: Thats fine. But what will be the value shown when a user fetches the metrics. Will it be shown as zero, or some other random float value.

blockdevice/blockdevice.go
```diff
@@ -282,6 +306,12 @@ type TemperatureInformation struct {
 
 	// CurrentTemperature is the temperature of the drive in celsius
 	CurrentTemperature int16
+
+	// LowestTemperature is the lowest temperature of the drive recorded in celsius
+	LowestTemperature int16
+
+	// HighestTemperature is the highest temperature of the drive recorded in celsius
+	HighestTemperature int16
```
akhilerm, line 314: Seachest also reports whether the reported highest/lowest temperature value is valid or not. Those fields also should be included as part of the metrics.

pkg/seachest/seachest.go
richardelling: per the SCSI spec, medium rotation rate = 0 means it is not reported. medium rotation rate = 1 means non-rotating medium (eg solid state)

ndm-exporter/collector/seachest.go
akhilerm: These are the commented out code
1. Rotation rate related fields [here](https://github.com/openebs/node-disk-manager/blob/d866a80652c63c1cba5cc354137caeaedb3418f7/cmd/ndm_daemonset/probe/seachestprobe.go#L125)
2. Temperature Data related fields [here](https://github.com/openebs/node-disk-manager/blob/d866a80652c63c1cba5cc354137caeaedb3418f7/cmd/ndm_daemonset/probe/seachestprobe.go#L164)

You need to remove that also, then only the metrics will be fetched.

pkg/seachest/seachest.go
akhilerm: @richardelling Iam a little confused

>  medium rotation rate = 1 means non-rotating medium (eg solid state)

Does rotation rate =1 means the device is an SSD?

harshthakur9030: @akhilerm as far as I understand, if we have an SSD, the latency reported will be zero. In the help section of the exported metrics, we've stated this part. 

@richardelling as I am only working with greater than 1, this code would be displaying latency for HDD, right?

pkg/metrics/smart/metrics.go
harshthakur9030: Didn't get it @akhilerm . Which way are they misaligned? Already ran go fmt on this.

pkg/metrics/smart/metrics.go
harshthakur9030: Please check out the section "Performance by disk size". https://cloud.google.com/compute/docs/disks/performance .

ndm-exporter/collector/seachest.go
harshthakur9030: The code that is commented is responsible for filling in the block device struct of NDM.  We're fetching the details from seachest package by giving the block device path and exporting those details with the prometheus exporter, it should work without uncommenting too but was there a reason it was commented in the first place(guessing performance issues and waiting for periodic fetching thing)?

blockdevice/blockdevice.go
harshthakur9030: Will add that. A bit confused about what does validity means here though. Is it whether that's applicable to the block device or something else?

akhilerm: To the blockdevice, every temperate metric that seachest gives also has an associated field which tells whether the reported data is valid or not. We have already added that field to the current temperature metric. It need to be added to the highest and lowest temperature also.

ndm-exporter/collector/seachest.go
akhilerm: ohkay. got it. Thanks.

> was there a reason it was commented in the first place

It was removed because the blockdevice struct did not have those fields. Earlier it was used when the disk resource was available. When moved to blockdevice resource all fields were not directly ported. Fields that were useful only were added as the need arises.

harshthakur9030: Oh ok. We can populate the block device struct with seachest data but is there a use case for that?

pkg/metrics/smart/metrics.go
akhilerm: The above is given for different sized SSDs to match performance  of a 7200 RPM drive. The performance comparison given in there and the actual RPM are different contexts. 

Drive RPM is something that you *cannot* change on the fly, even with cloud volumes. Its a hardware set value. And AFAIK seachest will not report this value for cloud volumes because they are virtual drives. It can be obtained only for physical disks.

pkg/metrics/smart/metrics.go
akhilerm: nvm, I thought `Set(` and `).` from line 341, should be at same level of indentation.

ndm-exporter/collector/seachest.go
```diff
@@ -185,8 +211,30 @@ func (sc *SeachestMetricData) getSeachestData() error {
...
+	klog.V(4).Infof("Rotation rate is %v", sc.RotationRate)
+	klog.V(4).Infof("Rotational latency is %v", sc.RotationalLatency)
+	klog.V(4).Infof("Current temperature is %v", sc.TempInfo.CurrentTemperature)
+	klog.V(4).Infof("Lowest temperature is %v", sc.TempInfo.LowestTemperature)
+	klog.V(4).Infof("Highest temperature is %v", sc.TempInfo.HighestTemperature)
+	klog.V(4).Infof("Capacity is %v", sc.Capacity)
+	klog.V(4).Infof("Total bytes read is %v", sc.TotalBytesRead)
+	klog.V(4).Infof("Total bytes written is %v", sc.TotalBytesWritten)
+	klog.V(4).Infof("Device utilization rate is %v", sc.DeviceUtilization)
+	klog.V(4).Infof("Endurance used is %v", sc.PercentEnduranceUsed)
```
utkarshmani1997, line 237: nit: you can use identifier +v to print fields of sc. If its pointer you need to implement String() method

utkarshmani1997 (dismissed):

pkg/metrics/smart/metrics.go
```diff
@@ -110,7 +175,99 @@ func (m *Metrics) WithBlockDeviceCurrentTemperatureValid() *Metrics {
...
+	return m
+}
+
+// WithBlockDeviceUtilizationRate declares the utilization rate of a block device
+func (m *Metrics) WithBlockDeviceUtilizationRate() *Metrics {
+	m.blockDeviceUtilizationRate = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: m.CollectorType,
+			Name:      "block_device_utilization_rate",
+			Help:      `utilization rate of a block device `,
```
richardelling, line 255: This is the utilization rate as a ratio of the vendor's expected workload wear factors. Values greater than 100% indicate the device has exceeded its designed workload. Can we change the help to be something like "ratio of actual workload vs manufacturer's designed workload for the device" ?

pkg/metrics/smart/metrics.go
```diff
@@ -110,7 +175,99 @@ func (m *Metrics) WithBlockDeviceCurrentTemperatureValid() *Metrics {
...
+	)
+	return m
+}
+
+// WithBlockDeviceUtilizationRate declares the utilization rate of a block device
+func (m *Metrics) WithBlockDeviceUtilizationRate() *Metrics {
+	m.blockDeviceUtilizationRate = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: m.CollectorType,
+			Name:      "block_device_utilization_rate",
```
richardelling, line 254: how about "block_device_utilization_rate_percent"?

pkg/metrics/smart/metrics.go
```diff
@@ -110,7 +175,99 @@ func (m *Metrics) WithBlockDeviceCurrentTemperatureValid() *Metrics {
...
+	)
+	return m
+}
+
+// WithBlockDevicePercentEnduranceUsed declares the percentage of endurance used by a block device
+func (m *Metrics) WithBlockDevicePercentEnduranceUsed() *Metrics {
+	m.blockDevicePercentEnduranceUsed = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: m.CollectorType,
+			Name:      "block_device_endurance_used",
```
richardelling, line 267: How about "block_device_endurance_used_percent"?

pkg/metrics/smart/metrics.go
```diff
@@ -110,7 +175,99 @@ func (m *Metrics) WithBlockDeviceCurrentTemperatureValid() *Metrics {
...
+			Help:      `Percent endurance used of a block device `,
```
richardelling, line 268: A better description is "estimate of the percentage of the device life that has been used"

pkg/seachest/seachest.go
richardelling: correct

ndm-exporter/collector/seachest.go
harshthakur9030: If we do that, we won't be able to give a description of the field during logging, right?

pkg/metrics/smart/metrics.go
```diff
@@ -31,6 +31,33 @@ type MetricsData struct {
...
+	blockDeviceRotationRate *prometheus.GaugeVec
+
+	// blockDeviceTotalReadBytes is the total number of bytes read from the block device
+	blockDeviceTotalReadBytes *prometheus.GaugeVec
+
+	// blockDeviceTotalWrittenBytes is the total number of bytes written from the block device
+	blockDeviceTotalWrittenBytes *prometheus.GaugeVec
+
+	// blockDeviceUtilizationRate is utilization rate of the block device
+	blockDeviceUtilizationRate *prometheus.GaugeVec
```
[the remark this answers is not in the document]
harshthakur9030: Nope, tested it only on GKE though.

pkg/metrics/smart/metrics.go
```diff
@@ -110,7 +175,99 @@ func (m *Metrics) WithBlockDeviceCurrentTemperatureValid() *Metrics {
...
+	return m
+}
+
+// WithBlockDeviceRotationalLatency declares the metric rotational latency
+// as a prometheus metric
+func (m *Metrics) WithBlockDeviceRotationalLatency() *Metrics {
+	m.blockDeviceRotationalLatency = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: m.CollectorType,
+			Name:      "block_device_rotational_latency",
```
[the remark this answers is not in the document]
harshthakur9030: I am wondering if units of the quantity better fit in the help section instead of the name. What do you think?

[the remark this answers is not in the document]
richardelling: The prometheus recommendations are to include the units in the metric name. https://prometheus.io/docs/practices/naming/
Don't worry about the length of the metric name, it doesn't really impact the work done in the database and humans tend to use the metrics through regex matching and don't actually have to type the full metric name.

pkg/metrics/smart/metrics.go
harshthakur9030: That would also mean rotational latency is constant. Is it okay if we move these two fields when we raise a PR for static exporter? Maybe comment it out for now?

ndm-exporter/collector/seachest.go
akhilerm: Yes, The blockdevice struct can be populated. The use case is, we need it for fetching the metrics.

pkg/metrics/smart/metrics.go
akhilerm: > That would also mean rotational latency is constant.

Yes.

> Is it okay if we move these two fields when we raise a PR for static exporter?

you can remove those fields, no need to keep them commented. and move it to the static exporter.

blockdevice/blockdevice.go
akhilerm: @harshthakur9030  PTAL.

harshthakur9030: Added lowest and highest temperature valid.

richardelling (approved):

cmd/ndm_daemonset/probe/seachestprobe.go
```diff
@@ -124,65 +124,65 @@ func (scp *seachestProbe) FillBlockDeviceDetails(blockDevice *blockdevice.BlockD
...
+	if blockDevice.SMARTInfo.TemperatureInfo.CurrentTemperatureDataValid == true {
+		blockDevice.SMARTInfo.TemperatureInfo.CurrentTemperature = seachestProbe.
 			SeachestIdentifier.GetCurrentTemperature(driveInfo)
 
 		klog.V(4).Infof("Disk: %s CurrentTemperature:%d filled by seachest.",
-			blockDevice.DevPath, blockDevice.TemperatureInfo.CurrentTemperature)
+			blockDevice.DevPath, blockDevice.SMARTInfo.TemperatureInfo.CurrentTemperature)
 
-		/*d.TemperatureInfo.HighestValid = seachestProbe.
+		blockDevice.SMARTInfo.TemperatureInfo.HighestTemperatureDataValid = seachestProbe.
```
akhilerm, line 164: Can the check be made for highest and lowest temperature data also. i.e fill the highest/lowest temp only if the corresponding field is valid. @harshthakur9030

akhilerm (changes requested):
Given one more suggestion on fetching data in seachest probe.

akhilerm (approved):
/lgtm

kmova (approved):
good one!

PR #463, opened 2020-07-22
fix(cleanup): run partprobe commands only on block files

akhilerm:

After cleaning up the device, partprobe was run on the device to reload the changes into the kernel. But running partprobe on sparse blockdevices can error out. 

**What this PR does?**:
-  Added an additional check to run `partprobe` only if the device is a block file.

No

Yes. The following 2 cases were tested.
1. Run the cleanup job on a sparse file
2. Run the cleanup job on a regular disk

Fix bug introduced in  #445 

**Checklist:**
- [x] Fixes [#3084](https://github.com/openebs/openebs/issues/3084)
- [ ] Commit has unit tests

### DIFF
--- a/pkg/cleaner/jobcontroller.go
+++ b/pkg/cleaner/jobcontroller.go
@@ -77,14 +77,37 @@ func NewCleanupJob(bd *v1alpha1.BlockDevice, volMode VolumeMode, tolerations []v
 		jobContainer.Command = []string{"/bin/sh", "-c"}
 
 		// fdisk is used to get all the partitions of the device.
+		// Example
+		// $ fdisk -o Device -l /dev/sda
+		// 	Disk /dev/sda: 465.8 GiB, 500107862016 bytes, 976773168 sectors
+		// 	Units: sectors of 1 * 512 = 512 bytes
+		// 	Sector size (logical/physical): 512 bytes / 4096 bytes
+		// 	I/O size (minimum/optimal): 4096 bytes / 4096 bytes
+		// 	Disklabel type: dos
+		// 	Disk identifier: 0x065e2357
+		//
+		// 	Device
+		// 	/dev/sda1
+		// 	/dev/sda2
+		// 	/dev/sda5
+		// 	/dev/sda6
+		// 	/dev/sda7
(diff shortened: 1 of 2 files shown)


pkg/cleaner/jobcontroller.go
```diff
@@ -83,8 +83,12 @@ func NewCleanupJob(bd *v1alpha1.BlockDevice, volMode VolumeMode, tolerations []v
 		// wipefs erases the filesystem signature from the block
 		// -a    wipe all magic strings
 		// -f    force erasure
-		args := fmt.Sprintf("(fdisk -o Device -l %s | grep \"^%s\" | xargs -I '{}' wipefs -fa '{}') && wipefs -fa %s && partprobe %s",
-			bd.Spec.Path, bd.Spec.Path, bd.Spec.Path, bd.Spec.Path)
+		args := fmt.Sprintf("(fdisk -o Device -l %[1]s "+
```
kmova, line 86: Is there any way we can add unit tests to verify that the grep expressions actually is accurate. 
an idea could be to simulate different outputs passed to the grep command ... compare the result of the grep command.

akhilerm: Comparing results means, we will have to execute the command right?

akhilerm: We can write a normal unit test case for each of the different types of blockdevices and the volume mode. But in these tests only `container.args` will be compared with the expected output. 

I think this is not what @kmova want?

kmova: How do we ensure/validate that the following lines are picking the right set of devices in different use cases?

```
fdisk -o Device -l %[1]s | grep \"^%[1]s\"
```

kmova: looks like a bit of magic sauce in there. probably comments to explain what each subcommand does also might increase some confidence.

akhilerm: Added the comments.

akhilerm:
@kmova PTAL, all the required comments have been added.

PR #469, opened 2020-08-04
fix(cleanup): fix cleanup job to run partprobe command only on disk types.

akhilerm:
cleanup job fails on partitions, because the partprobe command fails with invalid argument error.

**What this PR does?**:
Fix the cleanup script to run partprobe only on blockdevices having deviceType as disk

No

- claimed a sparse blockdevice, unclaimed and wiped it 
- claimed a whole disk for cstor, unclaimed and wiped it
- claimed a single partition for cstor, unclaimed and cleaned it up.
- claimed an LVM device for cstor, unclaimed and wiped it.

Continuation of #463 
Various scenarios tried out before arriving at this conclusion:
- Use ubuntu image instead of alpine base image for linux-utils container. This is because the partprobe implementation is different. But both partprobe implementations failed while running on partitions.
- Tried using the parent device name always for running partprobe. It always works. But getting parent name is difficult in operator
- Run partprobe only when deviceType is disk. (Current solution). The deviceType will be checked and partprobe will be run only if the deviceType is of disk.

Additional Information:
When partprobe was run on partition directly it failed. But when the partition contained a filesystem. it did not error out. The actual reason for this behaviour is not yet known. system calls were traced in both cases, but couldn't makeup any useful information out of it.

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/pkg/cleaner/jobcontroller.go
+++ b/pkg/cleaner/jobcontroller.go
@@ -105,9 +105,13 @@ func NewCleanupJob(bd *v1alpha1.BlockDevice, volMode VolumeMode, tolerations []v
 		args := fmt.Sprintf("(fdisk -o Device -l %[1]s "+
 			"| grep \"^%[1]s\" "+
 			"| xargs -I '{}' wipefs -fa '{}') "+
-			"&& wipefs -fa %[1]s "+
-			"&& if [ -b %[1]s ]; then partprobe %[1]s; fi;",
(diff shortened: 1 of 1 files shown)


akhilerm:
> did we test the lvm partition also? Claimed that for cstor and did the cleanup?

yes. @pawanpraka1  Tested with partitions on lvm also.

PR #492, opened 2020-09-29
fix(mount): detect real device when using /dev/root

zlymeda:
Some linux systems use device /dev/root as root filesystem, but /dev/root is not a real device. Thus symlink /sys/class/block/root does not exists and os-disk-exclude-filter fails to exclude os disk. If the symlink /sys/class/block/root does not exists, the real root disk is resolved from /proc/cmdline.

**What this PR does?**:
Detects real device behind /dev/root

No

* Spin up AWS EC2 (eg. m5d.xlarge using instance stores) with ubuntu 20.04
* install microk8s
* install helm
* install openebs using helm
* check the NDM logs:
  * there is an error saying "/sys/class/block/root No such file or directory"
  * the os disk is not filtered by os-disk-exclude-filter and bd is created
* using this change; the /dev/root is resolved via /proc/cmdline

**Checklist:**
- [ ] Fixes #<issue number>
- [x] Commit has unit tests

### DIFF
--- a/pkg/mount/mountutil.go
+++ b/pkg/mount/mountutil.go
@@ -27,6 +27,9 @@ import (
 
 var ErrCouldNotFindRootDevice = fmt.Errorf("could not find root device")
 
+const ProcCmdLine = "/proc/cmdline"
+const HostProcCmdLine = "/host/" + ProcCmdLine
+
 // DiskMountUtil contains the mountfile path, devpath/mountpoint which can be used to
 // detect partition of a mountpoint or mountpoint of a partition.
 type DiskMountUtil struct {
@@ -271,11 +274,10 @@ func (m *DiskMountUtil) getMountName(mountLine string) (DeviceMountAttr, bool) {
 }
 
 func getCmdlineFile() string {
-	cmdlineFile := "/host/proc/cmdline"
-	if !fileExists(cmdlineFile) {
-		cmdlineFile = "/proc/cmdline"
+	if fileExists(HostProcCmdLine) {
+		return HostProcCmdLine
 	}
-	return cmdlineFile
+	return ProcCmdLine
 }
 
 func getDeviceName(devPath string) string {
--- a/pkg/mount/mountutil.go
+++ b/pkg/mount/mountutil.go
@@ -27,8 +27,10 @@ import (
 
 var ErrCouldNotFindRootDevice = fmt.Errorf("could not find root device")
 
-const ProcCmdLine = "/proc/cmdline"
-const HostProcCmdLine = "/host/" + ProcCmdLine
+const (
+	procCmdLine     = "/proc/cmdline"
+	hostProcCmdLine = "/host" + procCmdLine
+)
 
 // DiskMountUtil contains the mountfile path, devpath/mountpoint which can be used to
 // detect partition of a mountpoint or mountpoint of a partition.
@@ -274,10 +276,10 @@ func (m *DiskMountUtil) getMountName(mountLine string) (DeviceMountAttr, bool) {
 }
 
 func getCmdlineFile() string {
-	if fileExists(HostProcCmdLine) {
-		return HostProcCmdLine
+	if fileExists(hostProcCmdLine) {
+		return hostProcCmdLine
 	}
-	return ProcCmdLine
+	return procCmdLine
 }
 
 func getDeviceName(devPath string) string {
--- a/pkg/mount/mountutil.go
+++ b/pkg/mount/mountutil.go
@@ -98,8 +98,11 @@ func (m DiskMountUtil) getDeviceMountAttr(fn getMountData) (DeviceMountAttr, err
 //	getDiskSysPath takes disk/partition name as input (sda, sda1, sdb, sdb2 ...) and
 //	returns syspath of that disk from which we can generate ndm given uuid of that disk.
 func getDiskDevPath(partition string) (string, error) {
-	// dev path be like /dev/sda4 we need to remove /dev/ from this string to get sys block path.
-	softlink := "/sys/class/block/" + partition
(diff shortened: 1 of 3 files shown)


pkg/mount/mountutil.go
```diff
@@ -112,6 +115,77 @@ func getDiskDevPath(partition string) (string, error) {
...
+
+//	getLinkForPartition returns path to sys block path
+func getLinkForPartition(partition string) string {
+	// dev path be like /dev/sda4 we need to remove /dev/ from this string to get sys block path.
+	return "/sys/class/block/" + partition
+}
+
+//	getRootPartition resolves link /dev/root using /proc/cmdline
+func getRootPartition() (string, error) {
+	file, err := os.Open("/proc/cmdline")
```
akhilerm, line 142: This will take `/proc` from the container right? We have the host `procfs` mounted at `/host/proc`. 
Should we use `/host/proc/cmdline` here instead?

pkg/mount/mountutil.go
```diff
@@ -112,6 +115,77 @@ func getDiskDevPath(partition string) (string, error) {
 	return "/dev/" + parentDisk, nil
 }
 
+//	getSoftLinkForPartition returns path to /sys/class/block/$partition
+//	if the path does not exist and the partition is "root"
+//	then the root partition is detected from /proc/cmdline
+func getSoftLinkForPartition(partition string) (string, error) {
+	softlink := getLinkForPartition(partition)
+
+	if _, err := os.Stat(softlink); os.IsNotExist(err) && partition == "root" {
```
akhilerm, line 124: question: Are there cases where `/dev/root` is symlink to the root partition?

pkg/mount/mountutil.go
zlymeda: yes, that's true. I checked the file inside of a container and it had the same content.
I can use /host/proc/cmdline and if it does not exists; I can check /proc/cmdline as a fallback (if not running in a container).
Is there any use-case when NDM would not run inside of a container?

pkg/mount/mountutil.go
zlymeda:
```shell
$ readlink -f /dev/root
/dev/root
```

it is not like a regular symlink. if the /dev/root is a real device, then /sys/class/block/root will exist and the original code will execute. the fallback checks if /sys/class/block/root does not exists.

pkg/mount/mountutil.go
akhilerm: > I can use /host/proc/cmdline and if it does not exists; I can check /proc/cmdline as a fallback (if not running in a container).

That would be great.



> Is there any use-case when NDM would not run inside of a container?

Nothing that I can think of now. cc: @kmova

pkg/mount/mountutil.go
```diff
@@ -112,6 +115,77 @@ func getDiskDevPath(partition string) (string, error) {
...
+			if !strings.HasPrefix(arg, rootPrefix) {
+				continue
+			}
+
+			rootSpec := strings.Split(arg[len(rootPrefix):], "=")
+
+			identifierType := strings.ToLower(rootSpec[0])
+			identifier := rootSpec[1]
+
+			path := fmt.Sprintf("/dev/disk/by-%s/%s", identifierType, identifier)
```
akhilerm, line 175: :+1:

pkg/mount/mountutil.go
akhilerm: I was referring to this answer. Though I am also not sure, whether it will be a symlink in any kernel version.

pkg/mount/mountutil.go
zlymeda: support for /host/proc/cmdline and using /proc/cmdline as fallback => implemented

pkg/mount/mountutil.go
```diff
@@ -185,3 +269,23 @@ func (m *DiskMountUtil) getMountName(mountLine string) (DeviceMountAttr, bool) {
 	}
 	return mountAttr, isValid
 }
+
+func getCmdlineFile() string {
+	cmdlineFile := "/host/proc/cmdline"
```
akhilerm, line 274: nit: Would be good to define `/host/proc/cmdline` and `/proc/cmdline` as a constant and then use it here.

akhilerm (commented):
One more small suggestion, everything else looks good.

pkg/mount/mountutil.go
zlymeda: done

pkg/mount/mountutil.go
```diff
@@ -19,11 +19,17 @@ package mount
...
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
 )
 
+var ErrCouldNotFindRootDevice = fmt.Errorf("could not find root device")
+
+const ProcCmdLine = "/proc/cmdline"
+const HostProcCmdLine = "/host/" + ProcCmdLine
```
akhilerm, line 31:
```suggestion
const procCmdLine = "/proc/cmdline"
const hostProcCmdLine = "/host/" + procCmdLine
```
Lets keep both of them unexported for now. WDYT?

Also, should we change it to 
```"/host" + procCmdLine```, else the resulting path will be `/host//proc/cmdline`

zlymeda: oh good catch

zlymeda: done

akhilerm (approved):
LGTM
Thanks for this fix @zlymeda .

kmova (approved):
Awesome! Thank you for this PR @zlymeda

kmova:
You can sign up for Hacktoberfest swag from MayaData by entering details here: https://github.com/mayadata-io/swag/issues/new/choose

PR #493, opened 2020-10-02
chore(CRD): Upgrade CRD version and add validation

RealHarshThakur:

## Pull Request template

Upgrades CRD version and adds validation. 
https://github.com/openebs/openebs/issues/3054

**What this PR does?**:
Upgrades CRD version to prevent deprecation of CRDs in future k8s release and adds OpenAPI validation. 

Currently, the new CRDs generated won't be applied to k8s. Ndm-operator still uses the old CRDs. 
In future, we will need to upgrade operator-sdk version : https://github.com/operator-framework/operator-sdk/issues/3558

**Checklist:**
- [x] Fixes https://github.com/openebs/openebs/issues/3054
- [ ] Commit has unit tests


deploy/crds/openebs.io_blockdeviceclaims_crd.yaml
```diff
@@ -0,0 +1,159 @@
...
+                      block device or a device that is a partition
+                    type: boolean
+                  blockVolumeMode:
+                    description: 'BlockVolumeMode represents whether to claim a device
+                      in Block mode or Filesystem mode. These are use cases of BlockVolumeMode:
+                      1) Not specified: VolumeMode check will not be effective 2)
+                      VolumeModeBlock: BD should not have any filesystem or mountpoint
+                      3) VolumeModeFileSystem: BD should have a filesystem and mountpoint.
+                      If DeviceFormat is    specified then the format should match
+                      with the FSType in BD'
```
akhilerm, line 66: Is the `description` generated from comments ?

pkg/apis/openebs/v1alpha1/blockdevice_types.go
```diff
@@ -39,21 +39,24 @@ type BlockDevice struct {
...
-	NodeAttributes NodeAttribute `json:"nodeAttributes"`
 
-	// Path contain devpath (e.g. /dev/sdb)
-	Path string `json:"path"`
+	// AggregateDevice was intended to store the hierachical
+	// information in cases of LVM. However this is currently
+	// not implemented and may need to be re-looked into for
+	// better design.
+	// +optional
+	AggregateDevice string `json:"aggregateDevice,omitempty"`
```
akhilerm, line 48: This field is deprecated. Can you add into the comments that this field is deprecated?

pkg/apis/openebs/v1alpha1/blockdevice_types.go
```diff
@@ -62,25 +65,26 @@ type DeviceSpec struct {
...
 
 	// ParentDevice was intended to store the UUID of the parent
 	// Block Device as is the case for partitioned block devices.
 	//
 	// For example:
-	// /dev/sda is the parent for /dev/sdap1
-	// TODO @kmova to be implemented/deprecated
+	// /dev/sda is the parent for /dev/sda1
+	// +kubebuilder:validation:Pattern:`^/dev/[a-z]{3,4}$`
 	ParentDevice string `json:"parentDevice,omitempty"`
```
akhilerm, line 77: This field is also deprecated.

pkg/apis/openebs/v1alpha1/blockdevice_types.go
```diff
@@ -62,25 +65,26 @@ type DeviceSpec struct {
...
-	// AggregateDevice was intended to store the hierachical
-	// information in cases of LVM. However this is currently
-	// not implemented and may need to be re-looked into for
-	// better design.
-	// TODO @kmova to be implemented/deprecated
-	AggregateDevice string `json:"aggregateDevice,omitempty"`
+	// Partitioned represents if BlockDevice has partitions or not (Yes/No)
+	// Currently always default to No.
+	// +kubebuilder:validation:Enum:=Yes;No
+	Partitioned string `json:"partitioned"`
```
akhilerm, line 82: This field is also deprecated.

pkg/apis/openebs/v1alpha1/blockdeviceclaim_types.go
```diff
@@ -50,7 +50,7 @@ type DeviceClaimSpec struct {
 	DeviceType string `json:"deviceType"`
 
 	// Node name from where blockdevice has to be claimed.
-	// TODO @akhilerm to be deprecated. Use NodeAttributes.HostName instead
+	// To be deprecated. Use NodeAttributes.HostName instead
 	HostName string `json:"hostName"`
```
akhilerm, line 54: :+1:

akhilerm (commented):
@harshthakur9030 Given some comments with regards to deprecated fields.

deploy/crds/openebs.io_blockdeviceclaims_crd.yaml
harshthakur9030: Yes, that's why I removed the //TODO comments.

harshthakur9030:
@akhilerm We cannot add deprecation to particular fields. When we create `v1` for the BD resource, we can just remove them without warning. We can only give deprecation warning that `v1alpha1` is deprecated and ask them to migrate to `v1` .

prateekpandey14 (commented):
`build/Dockerfile` file got added and content is not visible

harshthakur9030:
> `build/Dockerfile` file got added and content is not visible

@prateekpandey14 It's an empty file. Operator-sdk expects it to be there to generate CRDs

deploy/crds/openebs.io_blockdeviceclaims_crd.yaml
```diff
@@ -0,0 +1,159 @@
...
+kind: CustomResourceDefinition
+metadata:
+  name: blockdeviceclaims.openebs.io
+spec:
+  group: openebs.io
+  names:
+    kind: BlockDeviceClaim
+    listKind: BlockDeviceClaimList
+    plural: blockdeviceclaims
+    singular: blockdeviceclaim
```
akhilerm, line 11: shortName need to be added

deploy/crds/openebs.io_blockdeviceclaims_crd.yaml
```diff
@@ -0,0 +1,159 @@
...
+  scope: Namespaced
+  versions:
```
akhilerm, line 13: Additional printer columns need to be added.

deploy/crds/openebs.io_blockdeviceclaims_crd.yaml
harshthakur9030: @akhilerm Wasn't able to find the marker for it in kubebuilder doc. 
https://book.kubebuilder.io/reference/markers/crd-validation.html

harshthakur9030:
@akhilerm would I need to add `kubebuilder:validation:XPreserveUnknownFields` ? Are there any fields which NDM is adding and aren't present and set to optional in the CRD? 
https://book.kubebuilder.io/reference/markers/crd-processing.html

deploy/crds/openebs.io_blockdeviceclaims_crd.yaml
akhilerm: shortName is a field in the crd spec.

harshthakur9030: Yeah, but CRDs are generated by controller-gen, we shouldn't manually add it, I guess.

akhilerm: ooh. ohkay. How is it done in the other openebs CRDs?

akhilerm:
> @akhilerm would I need to add `kubebuilder:validation:XPreserveUnknownFields` ? Are there any fields which NDM is adding and aren't present and set to optional in the CRD?

I have to check through it. I am not excatly sure what it does. @prateekpandey14 Can you help here.

deploy/crds/openebs.io_blockdeviceclaims_crd.yaml
```diff
@@ -0,0 +1,159 @@
...
+                      VolumeModeBlock: BD should not have any filesystem or mountpoint
+                      3) VolumeModeFileSystem: BD should have a filesystem and mountpoint.
+                      If DeviceFormat is    specified then the format should match
+                      with the FSType in BD'
+                    type: string
+                  formatType:
+                    description: Format of the device required, eg:ext4, xfs
+                    type: string
+                type: object
+              deviceType:
```
akhilerm, line 72: this is an optional field

deploy/crds/openebs.io_blockdeviceclaims_crd.yaml
```diff
@@ -0,0 +1,159 @@
...
+                description: DeviceType represents the type of drive like SSD, HDD
+                  etc.,
+                type: string
+              hostName:
```
akhilerm, line 76: This is an optional field

deploy/crds/openebs.io_blockdeviceclaims_crd.yaml
```diff
@@ -0,0 +1,159 @@
...
+                      {key,value} in the matchLabels map is equivalent to an element
+                      of matchExpressions, whose key field is "key", the operator
+                      is "In", and the values array contains only "value". The requirements
+                      are ANDed.
+                    type: object
+                type: object
+            required:
+            - deviceType
+            - hostName
+            - resources
```
akhilerm, line 146: This is actually kind of conditional,

user can specify either the blockdevice name , or capacity request.

akhilerm:
@harshthakur9030 Can you remove the blank dockerfile? Though it may be required for code-gen I dont think it serves any purpose. Or is it mandatory to check in ?

deploy/crds/openebs.io_blockdevices_crd.yaml
```diff
@@ -0,0 +1,247 @@
...
+                      size in bytes
+                    format: int32
+                    type: integer
+                  storage:
+                    description: Storage is the blockdevice capacity in bytes
+                    format: int64
+                    type: integer
+                required:
+                - logicalSectorSize
+                - physicalSectorSize
```
akhilerm, line 59: These 2 fields are not mandatory.

deploy/crds/openebs.io_blockdevices_crd.yaml
```diff
@@ -0,0 +1,247 @@
...
+                    - loop
+                    - lvm
+                    - raid
+                    type: string
+                  driveType:
+                    description: DriveType is the type of backing drive, HDD/SSD
+                    enum:
+                    - HDD
+                    - SDD
+                    - Unknown
```
akhilerm, line 123: Should we consider this Unknown value, Is it similar to `""`?

akhilerm: Since we have given this field as a mandatory field, need to check if we should give this. Or we can remove it from mandatory field.

deploy/crds/openebs.io_blockdevices_crd.yaml
```diff
@@ -0,0 +1,247 @@
...
+                    type: string
+                  vendor:
+                    description: Vendor is vendor of disk
+                    type: string
+                required:
+                - deviceType
+                - driveType
+                - hardwareSectorSize
+                - logicalBlockSize
+                - physicalBlockSize
```
akhilerm, line 157: IMO, these fields should not be mandatory as there can cases where the blockdevices does not have these values.

deploy/crds/openebs.io_blockdevices_crd.yaml
```diff
@@ -0,0 +1,247 @@
...
+                  /dev/by-uuid/...
+                items:
+                  description: DeviceDevLink holds the mapping between type and links
+                    like by-id type or by-path type link
+                  properties:
+                    kind:
+                      description: Kind is the type of link like by-id or by-path.
+                      enum:
+                      - by-id
+                      - by-path
```
akhilerm, line 170: If I need to add a `by-uuid` or `by-partuuid` at a later time, should I modify the CRD also?

deploy/crds/openebs.io_blockdevices_crd.yaml
```diff
@@ -0,0 +1,247 @@
...
+                type: string
+              path:
+                description: Path contain devpath (e.g. /dev/sdb)
+                type: string
+            required:
+            - capacity
+            - details
+            - devlinks
+            - nodeAttributes
+            - partitioned
```
akhilerm, line 222: partitioned field is not mandatory.

deploy/crds/openebs.io_blockdevices_crd.yaml
```diff
@@ -0,0 +1,247 @@
...
+                enum:
+                - Claimed
+                - Unclaimed
+                - Released
+                type: string
+              state:
+                description: State is the current state of the blockdevice (Active/Inactive)
+                enum:
+                - Active
+                - Inactive
```
akhilerm, line 239: `Unknown` is also a valid state for a BlockDevice. When the NDM daemon on the node goes down, the blockdevices will be in an unknown state.

deploy/crds/openebs.io_blockdevices_crd.yaml
```diff
@@ -0,0 +1,247 @@
...
+                type: string
+            required:
+            - claimState
+            - state
```
akhilerm, line 243: :+1:

deploy/crds/openebs.io_blockdeviceclaims_crd.yaml
```diff
@@ -0,0 +1,159 @@
...
+                description: Resources will help with placing claims on Capacity,
+                  IOPS
+                properties:
+                  requests:
+                    additionalProperties:
+                      anyOf:
+                      - type: integer
+                      - type: string
+                      pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                      x-kubernetes-int-or-string: true
```
akhilerm, line 90: What does this field mean? We are expecting a `1Gi` kind of value right?

deploy/crds/openebs.io_blockdeviceclaims_crd.yaml
```diff
@@ -0,0 +1,159 @@
...
+                    description: 'Requests describes the minimum resources required.
+                      eg: if storage resource of 10G is requested minimum capacity
+                      of 10G should be available'
+                    type: object
+                required:
```
akhilerm, line 95: Is there a way to say the list of keys that are supported in the map, like we need only `storage` key in the map. All other keys should be invalid.

deploy/crds/openebs.io_blockdeviceclaims_crd.yaml
```diff
@@ -0,0 +1,159 @@
...
+                    description: matchLabels is a map of {key,value} pairs. A single
+                      {key,value} in the matchLabels map is equivalent to an element
+                      of matchExpressions, whose key field is "key", the operator
+                      is "In", and the values array contains only "value". The requirements
+                      are ANDed.
+                    type: object
+                type: object
+            required:
+            - deviceType
+            - hostName
```
akhilerm, line 145: hostName and deviceType is not mandatory.

deploy/crds/openebs.io_blockdeviceclaims_crd.yaml
harshthakur9030: Yep, it evaluates to any digit + any unit(kilo, mega, giga)

deploy/crds/openebs.io_blockdevices_crd.yaml
harshthakur9030: All of them?

deploy/crds/openebs.io_blockdevices_crd.yaml
harshthakur9030: `Unknown` is a string itself. Just thought that if we couldn't detect the drive type, we can mark it as unknown in the CRD.  
Yep, we can remove it from mandatory field

deploy/crds/openebs.io_blockdevices_crd.yaml
harshthakur9030: Yes, but I am not sure if same version of CR(`v1alpha1`) should have varying fields across NDM versions. Can we consider using metadata/annotations until we're confident about the fields that NDM needs?

deploy/crds/openebs.io_blockdevices_crd.yaml
akhilerm: yep. `hardwareSectorSize`, `logicalBlockSize`, `physicalBlockSize` are not mandatory.

deploy/crds/openebs.io_blockdeviceclaims_crd.yaml
akhilerm: @harshthakur9030 Can you check [this](https://github.com/openebs/api/blob/master/config/crds/all-crds.yaml). We should add the shortName.
Else `kubectl get bd` command wont work.

deploy/crds/openebs.io_blockdevices_crd.yaml
akhilerm: The fields are not varying here. We are not adding any new fields. Its about adding one more value that is supported. cc: @prateekpandey14

akhilerm:
@harshthakur9030 Can you push the latest changes to this PR?

harshthakur9030:
@akhilerm we're currently using operator-sdk version 0.19 . Shall I create an issue to migrate to v1 ?

akhilerm:
> Shall I create an issue to migrate to v1 ?

yes @harshthakur9030 . you can create an issue [here](https://github.com/openebs/openebs/issues)

pkg/apis/openebs/v1alpha1/blockdeviceclaim_types.go
```diff
@@ -29,6 +29,7 @@ import (
 // +k8s:openapi-gen=true
 
 // BlockDeviceClaim is the Schema for the BlockDeviceClaim CR
+// +kubebuilder:resource:scope=Namespaced,shortName=bd
```
akhilerm, line 32:
```suggestion
// +kubebuilder:resource:scope=Namespaced,shortName=bdc
```

pkg/apis/openebs/v1alpha1/blockdevice_types.go
```diff
@@ -28,7 +28,10 @@ import (
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +k8s:openapi-gen=true
 
+// +kubebuilder:resource:scope=Namespaced,shortName=bd
+
 // BlockDevice is the Schema used to represent a BlockDevice CR
+// +kubebuilder:resource:scope=Namespaced,shortName=bd
```
akhilerm, line 34: @harshthakur9030  Can you add the `additionalPrinterColumns`  also. ?

pkg/apis/openebs/v1alpha1/blockdeviceclaim_types.go
akhilerm: @harshthakur9030  Can you add the `additionalPrinterColumns`  also. ?

deploy/crds/openebs.io_blockdevices_crd.yaml
```diff
@@ -0,0 +1,238 @@
...
+                  deviceType:
+                    description: DeviceType represents the type of device like sparse,
+                      disk, partition, lvm, raid
+                    enum:
+                    - disk
+                    - partition
+                    - sparse
+                    - loop
+                    - lvm
+                    - raid
```
akhilerm, line 116:
```suggestion
                    - crypt
                    - dm
                    - mpath
```

deploy/crds/openebs.io_blockdeviceclaims_crd.yaml
```diff
@@ -0,0 +1,157 @@
...
+                type: string
+              hostName:
+                description: Node name from where blockdevice has to be claimed. To
+                  be deprecated. Use NodeAttributes.HostName instead
+                type: string
+              resources:
+                description: Resources will help with placing claims on Capacity,
+                  IOPS
+                properties:
+                  requests:
```
akhilerm, line 86: @harshthakur9030 . The following BDC still does not work.
```yaml
apiVersion: openebs.io/v1alpha1
kind: BlockDeviceClaim
metadata:
  name: bdc-ssd
spec:
  blockDeviceNodeAttributes:
    hostName: worker-ak1
  resources:
    requests:
    - storage: 1Gi 
```

with error
```
The BlockDeviceClaim "bdc-ssd" is invalid: spec.resources.requests: Invalid value: "array": spec.resources.requests in body must be of type object: "array"
```

deploy/crds/openebs.io_blockdevices_crd.yaml
akhilerm: Still hitting error of not accepting `""` empty value. I will make a change in code to change empty value to `Unknown`. But still I think we need to add `""` also here, otherwise the following case wont work.

1. Already a BD exists with empty value for driveType. And the users is trying to do some update operation like removing finalizer. This edit wont work, because the validation will fail for the `driveType` field.

akhilerm (changes requested):
@harshthakur9030 Some more changes are needed on this PR
1. additionalPrinterColumns for blockdevice and blockdeviceclaim CRD
2. driveType validation in blockdevice CRD

deploy/crds/openebs.io_blockdeviceclaims_crd.yaml
harshthakur9030: This is confusing. Any idea how to fix this?

prateekpandey14: i think its indentation problem it should be like this ?
```  yaml
resources:
    requests:
      storage: 1Gi
```

akhilerm: Sorry, It was en error in the yaml I had. Can ignore this comment.

pkg/apis/openebs/v1alpha1/blockdevice_types.go
harshthakur9030: @akhilerm I don't think it is required at CRD root level. I have provided it for other fields
https://book.kubebuilder.io/reference/markers/crd.html

deploy/crds/openebs.io_blockdevices_crd.yaml
```diff
@@ -13,7 +13,28 @@ spec:
...
+      name: Status
+      type: string
+    - jsonPath: .spec.path
+      name: Path
+      priority: 7
+      type: string
+    - jsonPath: .spec.filesystem.fsType
+      name: FSType
+      priority: 8
+      type: string
```
akhilerm, line 36: This needs to be reordered as specified here. https://github.com/openebs/node-disk-manager/blob/7e50f5b2e80fbf7efcff86b20f91b3e12686e687/pkg/setup/build_crd.go#L36.

Also, why is the priority set at 7 and 8. Just 1 would sufice.

deploy/crds/openebs.io_blockdevices_crd.yaml
```diff
@@ -238,3 +259,4 @@ spec:
         type: object
     served: true
     storage: true
+    subresources: {}
```
akhilerm, line 262: Why is this field added here? The subresource field is not required. It can be removed.

deploy/crds/openebs.io_blockdeviceclaims_crd.yaml
```diff
@@ -155,3 +165,4 @@ spec:
         type: object
     served: true
     storage: true
+    subresources: {}
```
akhilerm, line 168: This can be removed.

deploy/crds/openebs.io_blockdevices_crd.yaml
akhilerm: @harshthakur9030 Can you add empty value also here.

akhilerm:
@harshthakur9030 Can you fix the priority and ordering of printer columns. Also it would be good to remove the empty Dockerfile.

deploy/crds/openebs.io_blockdeviceclaims_crd.yaml
```diff
@@ -0,0 +1,168 @@
...
+  scope: Namespaced
+  versions:
+  - additionalPrinterColumns:
+    - jsonPath: .spec.blockDeviceName
+      name: BlockDeviceName
+      type: string
+    - jsonPath: .status.phase
+      name: Phase
+      type: string
+    - jsonPath: .metadata.creationTimestamps
```
akhilerm, line 23:
```suggestion
    - jsonPath: .metadata.creationTimestamp
```

akhilerm:
Got the following error from NDM operator pod, while trying to claim a blockdevice manually
```
E1203 06:24:04.351485       1 blockdeviceclaim_controller.go:122] bdc-ssd failed to claim: error updating status of BDC : bdc-ssd, BlockDeviceClaim.openebs.io "bdc-ssd" is invalid: spec.resources.requests: Invalid value: "null": spec.resources.requests in body must be of type object: "null"
```
The BDC used
```
apiVersion: openebs.io/v1alpha1
kind: BlockDeviceClaim
metadata:
  name: bdc-ssd
spec:
  blockDeviceName: blockdevice-19df294569e2c3ab97ebfd37b8d670c6
```

cc: @prateekpandey14 , @harshthakur9030

pkg/apis/openebs/v1alpha1/blockdeviceclaim_types.go
```diff
@@ -41,33 +45,41 @@ type BlockDeviceClaim struct {
 type DeviceClaimSpec struct {
 
 	// Selector is used to find block devices to be considered for claiming
+	// +optional
 	Selector *metav1.LabelSelector `json:"selector,omitempty"`
 
 	// Resources will help with placing claims on Capacity, IOPS
+	// +optional
```
prateekpandey14, line 52: add	`// +nullable` marker here since it can be null value

akhilerm (commented):
Changes looks good. @RealHarshThakur you need to generate the CRDs once again so that the latest changes in the `_types.go` are picked up.

akhilerm (approved):
Changes looks good. 
@prateekpandey14 Can you take a quick look

prateekpandey14 (approved):
lgtm

PR #500, opened 2020-10-08
feat(blockdevicefilter): add filter for empty block device tag label value

AJEETRAI707:
This PR adds a filter that does not select
BD that has empty block device tag value.
Ref: https://github.com/openebs/openebs/discussions/3139

Signed-off-by: ajeetrai707 <ajeetrai707@gmail.com>

## Pull Request template

**What this PR does?**:

**Checklist:**
- [ ] Fixes #<issue number>
- [x] Commit has unit tests

### DIFF
--- a/pkg/select/blockdevice/filters.go
+++ b/pkg/select/blockdevice/filters.go
@@ -265,13 +265,32 @@ func filterNodeName(originalBD *apis.BlockDeviceList, spec *apis.DeviceClaimSpec
 // already been filtered by the given selector.
 func filterBlockDeviceTag(originalBD *apis.BlockDeviceList, spec *apis.DeviceClaimSpec) *apis.BlockDeviceList {
 
+	/// remove BDs with empty tag in all cases
+	emptyTagFilteredBDList := &apis.BlockDeviceList{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "BlockDevice",
+			APIVersion: "openebs.io/v1alpha1",
+		},
+	}
+	// This is given as a separate filter loop that works in all cases, irrespective of whether the user
+	// has specified the the label selector in the BDC or not.
+	for _, bd := range originalBD.Items {
+		// if the tag label value is empty then the BD should not be selected
+		if val, ok := bd.Labels[kubernetes.BlockDeviceTagLabel]; ok {
+			if len(strings.TrimSpace(val)) == 0 {
+				continue
+			}
+		}
+		emptyTagFilteredBDList.Items = append(emptyTagFilteredBDList.Items, bd)
+	}
+
 	// if the block-device-tag label was already included in the selector
 	// given in the BDC by the user, then this filter is not required. This
 	// is because it would have already performed the filter operation with the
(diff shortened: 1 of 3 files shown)


pkg/select/blockdevice/filters.go
```diff
@@ -26,6 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/klog"
+	"strings"
```
akhilerm, line 29: Can you separate the imports like [this](https://github.com/openebs/node-disk-manager/blob/3e813577dd07ec3a65ca386feaf9d893024cd1da/pkg/partition/partition.go#L19)

pkg/select/blockdevice/filters.go
```diff
@@ -296,6 +297,12 @@ func filterBlockDeviceTag(originalBD *apis.BlockDeviceList, spec *apis.DeviceCla
 	}
 
 	for _, bd := range originalBD.Items {
+		// if the tag label value is empty then the BD should not be selected
```
akhilerm, line 300: Also, add unit test cases for this code block.

akhilerm (commented):
@AJEETRAI707 , given a few comments.

pkg/select/blockdevice/filters_test.go
```diff
@@ -164,6 +191,42 @@ func TestFilterBlockDeviceTag(t *testing.T) {
...
+		"some BDs with tag key, with selector matching empty tag": {
+			args: args{
+				bdLabelList: bdLabelList4,
+				spec: &apis.DeviceClaimSpec{
+					Selector: &v1.LabelSelector{
+						MatchLabels: map[string]string{kubernetes.BlockDeviceTagLabel: ""},
+					},
+				},
+			},
+			wantedNoofBDs: 4,
```
kmova, line 212: Also, is it allowed to pass a BD tag with empty value in BDC. 
  
Shouldn't this be 0?  because `""` will always be excluded.

kmova: cc: @akhilerm

akhilerm: > Also, is it allowed to pass a BD tag with empty value in BDC.

Yes. it is allowed, because its a valid label selector.

> Shouldn't this be 0? because "" will always be excluded.

Actually 2 levels of filtering happens. The original BD List that we get here, will be after a label selection as part of the list call. So those selectors will already be applied. But if those selectors are applied, we still need to remove the empty tag, thats being done in this filter.

So the tests assume that already a selection has happened with the list call. And the output of the tests will be therefore including the BDs that would have been removed by the select call. Thats why the additional 4 BDs come.

akhilerm: If you check [this](https://github.com/openebs/node-disk-manager/pull/500/files#diff-10b936d2720dd7e71a105e22758229744715db30e38ca881ed9b9f282fc044b7R288) comment , you can see how the filter works.

kmova (approved):

PR #503, opened 2020-10-18
chore(build) Updating dockerfile(s) with buildx built-in ARGs

xUnholy:
Signed-off-by: Michael Fornaro <20387402+xUnholy@users.noreply.github.com>

Updating Dockerfiles to use the `buildx` built-in ARGs will provide more flexibility to provide support for additional OS/ARCHS/ARM variants in the future. This also will create a cleaner and more maintainable Dockerfile moving forward.

https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests
- [ ] (Optional) If documentation changes are required, which issue on https://github.com/openebs/openebs-docs is used to track them

### DIFF
--- a/build/ndm-daemonset/Dockerfile
+++ b/build/ndm-daemonset/Dockerfile
@@ -16,10 +16,15 @@
 #
 FROM golang:1.14.7 as build
 
-ARG TARGETPLATFORM
+ARG TARGETOS
+ARG TARGETARCH
+ARG TARGETVARIANT=""
 
 ENV GO111MODULE=on \
   CGO_ENABLED=1 \
+  GOOS=${TARGETOS} \
+  GOARCH=${TARGETARCH} \
+  GOARM=${TARGETVARIANT} \
   DEBIAN_FRONTEND=noninteractive \
   PATH="/root/go/bin:${PATH}"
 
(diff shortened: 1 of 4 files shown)


build/ndm-daemonset/Dockerfile
```diff
@@ -16,10 +16,15 @@
 #
 FROM golang:1.14.7 as build
 
-ARG TARGETPLATFORM
+ARG TARGETOS
+ARG TARGETARCH
+ARG TARGETVARIANT=""
```
akhilerm, line 21: Why do we need to set the `TARGETVARIANT` as empty string?, while we dont set the other envs.

akhilerm (commented):
@xUnholy : given a comment / question.

build/ndm-daemonset/Dockerfile
xUnholy: This is because it is only set when it's given armv7 or 8 or equivalent, if we move any additional logic from this variable this provides just a default empty string rather than null value.

akhilerm (approved):
/lgtm

kmova (approved):

PR #517, opened 2020-12-04
fix(eventhandler): continue processing indepedent blockdevices even if one device fails

akhilerm:
When one of the disk is bad in the node / NDM is not able to create blockdevice resource for the disk, NDM repeatedly performs the rescan operation, skipping the other disks. This results in some disks not being processed. This PR brings in a change to wait till all the disks are processed, before starting the rescan.

**What this PR does?**:
- adds a check to skip processing of blockdevices if the parent or any of the slave devices of the given block device has errored out.

NOTE: This commit does not change the continuous rescan that NDM will do incase of a disk erroring out.

Solves one issue mentioned in https://github.com/openebs/openebs/issues/3051

**Checklist:**
- [ ] Fixes 
- [x] Commit has unit tests

### DIFF
--- a/cmd/ndm_daemonset/probe/eventhandler.go
+++ b/cmd/ndm_daemonset/probe/eventhandler.go
@@ -17,6 +17,7 @@ limitations under the License.
 package probe
 
 import (
+	"github.com/openebs/node-disk-manager/blockdevice"
 	"github.com/openebs/node-disk-manager/cmd/ndm_daemonset/controller"
 	"github.com/openebs/node-disk-manager/pkg/features"
 	libudevwrapper "github.com/openebs/node-disk-manager/pkg/udev"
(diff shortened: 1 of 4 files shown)


cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -62,12 +66,14 @@ func (pe *ProbeEvent) addBlockDeviceEvent(msg controller.EventMessage) {
 		klog.Infof("Processed details for %s", device.DevPath)
 
 		if isGPTBasedUUIDEnabled {
+			if isParentOrSlaveDevice(*device, erroredDevices) {
+				continue
```
kmova, line 70: log an error for this device as well.

kmova: a warning message that we skipped it because the device parent or child is already in an errored state.

akhilerm: Done.

kmova (approved):

PR #524, opened 2020-12-29
feat(filter): add support to use unused partitions in os-disk

akhilerm:
Sometimes the nodes may have only one large disk with many partitions. NDM currently ignores the complete disk if it contains a system partition. This PR adds support to create blockdevice resources from the unused partitions on the OS disk.

**What this PR does?**:
- adds a new feature gate to enable / disable using OS disk for blockdevices
- modify the os disk filter to exclude only the partition that contains the system mountpoints and not the entire disk

No

**Checklist:**
- [x] Fixes openebs/openebs#3133
- [x] Commit has unit tests

### DIFF
--- a/pkg/features/features.go
+++ b/pkg/features/features.go
@@ -41,8 +41,11 @@ const (
 	// blockdevice UUID algorithm mentioned in
 	// https://github.com/openebs/openebs/pull/2666
 	GPTBasedUUID Feature = "GPTBasedUUID"
+
 	// APIService feature flag starts the GRPC server which provides functionality to manage block devices
 	APIService Feature = "APIService"
+
+	UseOSDisk Feature = "UseOSDisk"
 )
 
 // supportedFeatures is the list of supported features. This is used while parsing the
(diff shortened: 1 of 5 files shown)


pkg/features/features.go
```diff
@@ -41,21 +41,26 @@ const (
...
 	GPTBasedUUID,
 	APIService,
+	UseOSDisk,
 }
 
 // defaultFeatureGates is the default features that will be applied to the application
 var defaultFeatureGates = map[Feature]bool{
 	GPTBasedUUID: false,
 	APIService:   false,
+	UseOSDisk:    false,
```
kmova, line 63: Should the GPTBasedUUID be `true` by default?

pkg/mount/mountutil.go
```diff
@@ -116,11 +117,22 @@ func getDiskDevPath(partition string) (string, error) {
 		return "", err
 	}
 
-	parentDisk, ok := getParentBlockDevice(link)
-	if !ok {
-		return "", fmt.Errorf("could not find parent device for %s", link)
+	var disk string
```
kmova, line 120: n_: Fix the function comment => `getDiskSysPath` to `getDiskDevPath` ==> Should this intead be called `getPartitionDevPath` ?

kmova: Similar to other functions include information on how this function will be used. This will help in reviewing the code added under the feature flag.

pkg/features/features.go
akhilerm: yes, @kmova . I have already made those changes in another branch. Dont want to mix it up with current PR. Currently we set it to default by using the flag by default while starting. Setting this to true will help us to avoid that.  In the other branch, I also have fixes for some tests once this is made default.

pkg/mount/mountutil.go
akhilerm: Done.

kmova (approved):

PR #531, opened 2021-01-13
[stable/ndm]: update ndm charts to 1.1.0

akhilerm:
- update chart version
- update README
- update values.yaml
- update NDM daemonset and NDM operator

Documentation change for upgrade #533 

### DIFF
--- a/deploy/helm/charts/Chart.yaml
+++ b/deploy/helm/charts/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: openebs-ndm
-description: Kubernetes Storage Device Management. Also used to maintain the inventory of block devices.
+description: Helm chart for OpenEBS Node Disk Manager - a Kubernetes native storage device management solution. For instructions on how to install, refer to https://openebs.github.io/node-disk-manager/.
 version: 1.1.0
 appVersion: 1.1.0
 icon: https://raw.githubusercontent.com/cncf/artwork/master/projects/openebs/icon/color/openebs-icon-color.png
--- a/deploy/helm/charts/Chart.yaml
+++ b/deploy/helm/charts/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v2
 name: openebs-ndm
 description: Kubernetes Storage Device Management. Also used to maintain the inventory of block devices.
-version: 1.0.2
-appVersion: 1.0.1
+version: 1.1.0
+appVersion: 1.1.0
 icon: https://raw.githubusercontent.com/cncf/artwork/master/projects/openebs/icon/color/openebs-icon-color.png
 home: http://www.openebs.io/
 keywords:
--- a/deploy/helm/charts/crds/blockdevice.yaml
+++ b/deploy/helm/charts/crds/blockdevice.yaml
@@ -28,12 +28,236 @@ spec:
       name: Age
       type: date
   group: openebs.io
-  version: v1alpha1
-  scope: Namespaced
   names:
     kind: BlockDevice
     listKind: BlockDeviceList
     plural: blockdevices
-    singular: blockdevice
     shortNames:
       - bd
+    singular: blockdevice
+  scope: Namespaced
+  subresources: {}
+  validation:
+    openAPIV3Schema:
+      description: BlockDevice is the Schema used to represent a BlockDevice CR
+      properties:
+        apiVersion:
+          description: 'APIVersion defines the versioned schema of this representation
+            of an object. Servers should convert recognized schemas to the latest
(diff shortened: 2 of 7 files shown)


deploy/helm/charts/templates/deployment.yaml
```diff
@@ -64,6 +64,10 @@ spec:
         # from NDM Operator. By default the CRDs will be installed
         - name: OPENEBS_IO_INSTALL_CRD
           value: "{{ .Values.crd.enableInstall }}"
+{{- if .Values.imagePullSecrets }}
```
kmova, line 67: 👍

deploy/helm/charts/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v2
 name: openebs-ndm
 description: Kubernetes Storage Device Management. Also used to maintain the inventory of block devices.
```
kmova, line 3: n_: Change this description to the following, as it gets added to the chart release that is made. 


`Helm chart for OpenEBS Node Disk Manager - a Kubernetes native storage device management solution. For instructions on how to install, refer to  https://openebs.github.io/node-disk-manager/.`

akhilerm: Done.

deploy/helm/charts/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v2
 name: openebs-ndm
-description: Kubernetes Storage Device Management. Also used to maintain the inventory of block devices.
-version: 1.0.2
-appVersion: 1.0.1
+description: Helm chart for OpenEBS Node Disk Manager - a Kubernetes native storage device management solution. For instructions on how to install, refer to https://openebs.github.io/node-disk-manager/.
+version: 1.1.0
+appVersion: 1.1.0
```
xUnholy, line 5: Is this the accurate version of the NDM application?

xUnholy: Just checked, seems to be correct 👍

deploy/helm/charts/crds/blockdevice.yaml
```diff
@@ -28,12 +28,236 @@ spec:
...
-  version: v1alpha1
-  scope: Namespaced
   names:
     kind: BlockDevice
     listKind: BlockDeviceList
     plural: blockdevices
-    singular: blockdevice
     shortNames:
       - bd
+    singular: blockdevice
```
xUnholy, line 37: > There is no support at this time for upgrading or deleting CRDs using Helm. This was an explicit decision after much community discussion due to the danger for unintentional data loss. Furthermore, there is currently no community consensus around how to handle CRDs and their lifecycle. As this evolves, Helm will add support for those use cases.

Does this mean we will need to provide some docs on removing the old CRD and replacing it with the new one perhaps?

akhilerm: Where was this mentioned @xUnholy . I think helm3 added support for CRDs. cc: @sonasingh46

kmova: I do remember reading about the above. 

Assuming the upgrade helm chart has `crd.enableInstall=true`, will the NDM operator go ahead and upgrade the CRDs with latest content?

xUnholy: This is from the Helm 3 docs @akhilerm . My understanding is this is usually a thing that causes problems and we need users to remove previously installed CRDs and to re-install these.

kmova: Removing a CRD will result in deleting the old CRs.. right? Or has that changed recently. 

Also, an alternative could be to ask users to first upgrade the CRDs by running a kubectl command, prior to upgrading the NDM helm chart. 

@xUnholy -- is there any other best practice to follow for handling the helm upgrades that do require changes to the CRDs?

xUnholy: I think if they uplift the CRDs prior that is probably the logical path, I haven't had heaps of experiences with managing CRDs in helm but they can cause these types of issues.

Would suggest we did some testing to see what is going to cause the least impact.

akhilerm: Tested the following case
1. Installed NDM charts with version 1.0.2, 
2. Claimed a BlockDevice
3. Upgraded the charts to latest version (with upgraded CRDs)
`helm upgrade openebs -n openebs .`

The CRDs were not upgraded and they were kept as in the old version.

akhilerm: @xUnholy @kmova Updated the upgrade steps in #533

xUnholy (approved):

PR #537, opened 2021-01-18
fix(probe): fetch block size and drive type for partitions

Z0Marlin:
sysfsprobe fetches logical/physical blocksize, hw sector size,
and drive type from sysfs using `<syspath>/queue` directory.
This directory is available only for parent disks and not partitions.
These details for partitions are same as their parent disks. Use
their parent disks to fetch the details.

## Testing

Manually tested by running in a minikube cluster
Screenshot from 2021-01-20 17-15-00

**Checklist:**
- [x] Fixes openebs/openebs#3134
- [ ] Commit has unit tests

### DIFF
--- a/cmd/ndm_daemonset/probe/sysfsprobe.go
+++ b/cmd/ndm_daemonset/probe/sysfsprobe.go
@@ -90,6 +90,33 @@ func (cp *sysfsProbe) FillBlockDeviceDetails(blockDevice *blockdevice.BlockDevic
 		return
 	}
 
+	if blockDevice.Capacity.Storage == 0 {
+		capacity, err := sysFsDevice.GetCapacityInBytes()
+		if err != nil {
+			klog.Warningf("unable to get capacity for device: %s, err: %v", blockDevice.DevPath, err)
+		}
+		blockDevice.Capacity.Storage = uint64(capacity)
+		klog.V(4).Infof("blockdevice path: %s capacity :%d filled by sysfs probe.",
+			blockDevice.DevPath, blockDevice.Capacity.Storage)
+	}
+
+	// If the blockdevice is a partition, we will use its parent disk to get block size, hw
+	// sector size and drive type.
+	// Get the parent disk sysfs device using the parent's dev path stored in the blokdevice
+	if blockDevice.DeviceAttributes.DeviceType == blockdevice.BlockDeviceTypePartition {
+		parentDev := blockDevice.DependentDevices.Parent
+		if parentDev == "" {
+			klog.Errorf("cannot find the parent disk for partition: %s", blockDevice.DevPath)
+			return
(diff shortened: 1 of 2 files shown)


cmd/ndm_daemonset/probe/sysfsprobe.go
```diff
@@ -89,9 +89,26 @@ func (cp *sysfsProbe) FillBlockDeviceDetails(blockDevice *blockdevice.BlockDevic
 		klog.Errorf("unable to get sysfs device for device: %s, err: %v", blockDevice.DevPath, err)
 		return
 	}
+	pSysFsDevice := sysFsDevice
```
akhilerm, line 92: Why is this assignment done?

cmd/ndm_daemonset/probe/sysfsprobe.go
```diff
@@ -115,7 +132,7 @@ func (cp *sysfsProbe) FillBlockDeviceDetails(blockDevice *blockdevice.BlockDevic
 	}
 
 	if blockDevice.DeviceAttributes.HardwareSectorSize == 0 {
-		hwSectorSize, err := sysFsDevice.GetPhysicalBlockSize()
+		hwSectorSize, err := pSysFsDevice.GetPhysicalBlockSize()
```
akhilerm, line 135:
```suggestion
		hwSectorSize, err := pSysFsDevice.GetHardwareSectorSize()
```

akhilerm (commented):
@Z0Marlin Have given some comments.

cmd/ndm_daemonset/probe/sysfsprobe.go
Z0Marlin: Because `sysFsDevice` is still being used to calculate the capacity here - https://github.com/openebs/node-disk-manager/blob/507409ecc94216ddc467a38e84c1899e4228da0c/cmd/ndm_daemonset/probe/sysfsprobe.go#L158
So a reference to the original blockdevice is still needed

Z0Marlin: another way would be to just move the call which requires a reference to the original blockdevice above all the other calls, and reassign `sysFsDevice` after that

akhilerm: > another way would be to move the call which requires a reference to the original blockdevice above all the other calls, and reassign sysFsDevice after that

@Z0Marlin , I thnk this method ^^^ will be better. Anything that requires parent device comes after reassining. Also can you rename `psysfsDevice` to `parentSysFSDevice` and the same with `pDev` to `parentDev`

Z0Marlin: done

PR #574, opened 2021-05-06
(refactor, bugsbash)fix defer close statements

asquare14:
Signed-off-by: aSquare14 <atibhi.a@gmail.com>

Related to bugsbash

### DIFF
--- a/pkg/mount/mountutil.go
+++ b/pkg/mount/mountutil.go
@@ -80,9 +80,17 @@ func (m DiskMountUtil) getDeviceMountAttr(fn getMountData) (DeviceMountAttr, err
 	if err != nil {
 		return mountAttr, err
 	}
+
 	defer func() {
-		file.Close()
+		cerr := file.Close()
+		if cerr != nil {
+			err = cerr
+		}
 	}()
+	if err != nil {
+		return mountAttr, err
+	}
+
 	scanner := bufio.NewScanner(file)
 	if err := scanner.Err(); err != nil {
 		return mountAttr, err
@@ -171,8 +179,12 @@ func getRootPartition() (string, error) {
 	if err != nil {
 		return "", err
 	}
+
 	defer func() {
-		file.Close()
+		cerr := file.Close()
+		if cerr != nil {
+			err = cerr
+		}
 	}()
 
 	path, err := parseRootDeviceLink(file)
@@ -185,7 +197,7 @@ func getRootPartition() (string, error) {
 		return "", err
 	}
 
-	return getDeviceName(link), nil
+	return getDeviceName(link), err
 }
 
 func parseRootDeviceLink(file io.Reader) (string, error) {
--- a/pkg/mount/mountutil.go
+++ b/pkg/mount/mountutil.go
@@ -19,11 +19,12 @@ package mount
 import (
 	"bufio"
 	"fmt"
-	"github.com/openebs/node-disk-manager/pkg/features"
 	"io"
 	"os"
 	"path/filepath"
 	"strings"
+
+	"github.com/openebs/node-disk-manager/pkg/features"
 )
 
 var ErrCouldNotFindRootDevice = fmt.Errorf("could not find root device")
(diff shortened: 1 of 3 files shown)


akhilerm:
@asquare14 Can you check on the comments by the muse-bot

asquare14:
@akhilerm Done ! Thanks for the review

asquare14:
Can you please show me a screenshot of Better Code Hub, I am unable to view it.

pkg/mount/mountutil.go
```diff
@@ -79,7 +80,17 @@ func (m DiskMountUtil) getDeviceMountAttr(fn getMountData) (DeviceMountAttr, err
 	if err != nil {
 		return mountAttr, err
 	}
-	defer file.Close()
+
+	defer func() {
```
akhilerm, line 84: The defer block will be executed only when the function is returned. So why is the `err` assignment done?

pkg/mount/mountutil.go
```diff
@@ -168,7 +179,13 @@ func getRootPartition() (string, error) {
 	if err != nil {
 		return "", err
 	}
-	defer file.Close()
+
+	defer func() {
+		cerr := file.Close()
+		if cerr != nil {
+			err = cerr
```
akhilerm, line 186: Same comment as above

pkg/mount/mountutil.go
```diff
@@ -180,7 +197,7 @@ func getRootPartition() (string, error) {
 		return "", err
 	}
 
-	return getDeviceName(link), nil
+	return getDeviceName(link), err
```
akhilerm, line 200: :+1:

pkg/udev/mockdata.go
```diff
@@ -25,12 +25,13 @@ package udev
...
-	bd "github.com/openebs/node-disk-manager/blockdevice"
-	"github.com/openebs/node-disk-manager/pkg/mount"
-	"github.com/openebs/node-disk-manager/pkg/sysfs"
 	"io/ioutil"
 	"os"
 	"strings"
+
+	bd "github.com/openebs/node-disk-manager/blockdevice"
+	"github.com/openebs/node-disk-manager/pkg/mount"
+	"github.com/openebs/node-disk-manager/pkg/sysfs"
```
akhilerm, line 34: :+1:

akhilerm (commented):
Given some comments.

akhilerm:
> Can you please show me a screenshot of Better Code Hub, I am unable to view it.

@asquare14 The Better code hub failure is expected. Its not because of changes in this PR. Only because the file was evaluated now, the issues from previous changes came in.

pkg/mount/mountutil.go
asquare14: Fixed

pkg/mount/mountutil.go
```diff
@@ -79,7 +80,15 @@ func (m DiskMountUtil) getDeviceMountAttr(fn getMountData) (DeviceMountAttr, err
...
 	}
-	defer file.Close()
+
+	defer func() {
+		file.Close()
+	}()
+
+	if err != nil {
+		return mountAttr, err
+	}
```
akhilerm, line 90: This error check is not needed here.

pkg/mount/mountutil.go
```diff
@@ -168,7 +177,10 @@ func getRootPartition() (string, error) {
 	if err != nil {
 		return "", err
 	}
-	defer file.Close()
+
+	defer func() {
+		file.Close()
+	}()
```
akhilerm, line 183: @asquare14 What is the advantage of using the above format for `defer` compared to
```golang
defer file.Close()
```

richardelling: file.Close() can fail and you need to check the return code. This is not a problem for files opened for reading, but can be  a big problem for files opened for writing. The automated code checkers aren't very smart about this, so they assume the worst: the file is open for writing and you need to deal with failed Close(). 

NB, the above code still ignores the error condition, but it fakes out the automated code checker. Perhaps a better solution is to either explicitly ignore the error condition in the defer func for files opened for read or add a comment so the code checker ignores that check for this instance.

akhilerm: @asquare14 Can you make the necessary changes as per the above comment. Here, the file is opened for reading. So,  we can explicitly ignore the error.

richardelling: what I mean by explicitly ignoring for files that are opened for reading is something like:
```
defer func() {
    _ := file.Close()
}
```

akhilerm:
@asquare14 Can you address the review comments on the PR.

PR #585, opened 2021-05-07
Removed weak random number generator by adding new import

pankaj892:
Signed-off-by: Pankaj Biradar <pankaj28988@gmail.com>

### DIFF
--- a/integration_tests/udev/disks.go
+++ b/integration_tests/udev/disks.go
@@ -19,7 +19,7 @@ package udev
 import (
 	"fmt"
 	"github.com/openebs/node-disk-manager/integration_tests/utils"
-	"math/rand"
+	"crypto/rand"
 	"os"
 	"strconv"
 	"time"
--- a/integration_tests/udev/disks.go
+++ b/integration_tests/udev/disks.go
@@ -19,7 +19,7 @@ package udev
 import (
 	"fmt"
 	"github.com/openebs/node-disk-manager/integration_tests/utils"
-	"math/rand" #nosec G404
+	"math/rand" // #nosec G404
 	"os"
 	"strconv"
 	"time"
--- a/integration_tests/udev/disks.go
+++ b/integration_tests/udev/disks.go
@@ -19,7 +19,7 @@ package udev
 import (
 	"fmt"
 	"github.com/openebs/node-disk-manager/integration_tests/utils"
-	"crypto/rand"
+	"math/rand" #nosec G404
 	"os"
 	"strconv"
 	"time"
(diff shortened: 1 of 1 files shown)


integration_tests/udev/disks.go
```diff
@@ -19,7 +19,7 @@ package udev
 import (
 	"fmt"
 	"github.com/openebs/node-disk-manager/integration_tests/utils"
-	"math/rand"
+	"crypto/rand"
```
akhilerm, line 22: The corresponding usages also needs to be changed as `crypto/rand` does not depend on a seed value.

akhilerm (changes requested):
Given some comments.

integration_tests/udev/disks.go
akhilerm: @pankaj892 Can you update on this PR?

pankaj892: @akhilerm So so i need to remove the seed method also or do i keep it as it is?

akhilerm: @pankaj892 This change of random number generation from `math/rand` to `crypto/rand` is actually not required here , since this is only part of the tests and cryptographic security is not needed. But still the change can be done. To know how to generate random numbers using crypto package refer [here](https://flaviocopes.com/go-random/#:~:text=Pseudorandom%20numbers,-The%20math%2Frand&text=As%20an%20example%2C%20try%20running,between%200%20and%20n%20%2D%201%20.&text=This%20is%20because%20by%20default,unique%20seed%20for%20your%20program.)

pankaj892: @akhilerm if i use crypto/rand then would it make the generation of random numbers slow or should i revert it back to math/rand?

richardelling: IMHO `math/rand` is fine, but you'll likely want to add a comment `//#nosec G404`  (or whatever the complaining check is)
We just don't need crypto-strength randoms for this and the code checkers default to being paranoid.

akhilerm: cc: @pankaj892 PTAL

pankaj892: @akhilerm  i have made the necessary changes please take a look and let me know if any changes needed

integration_tests/udev/disks.go
```diff
@@ -19,7 +19,7 @@ package udev
 import (
 	"fmt"
 	"github.com/openebs/node-disk-manager/integration_tests/utils"
-	"math/rand"
+	"math/rand" #nosec G404
```
akhilerm, line 22: This will give compile error. Need to replace with
```suggestion
	"math/rand" // #nosec G404
```

akhilerm (commented):
given a comment

pankaj892:
@akhilerm check if its ok now

akhilerm (approved):
LGTM

PR #590, opened 2021-05-10
fix(pkg, mount): fetch all mount points for a given device

### LINKED ISSUE #558: NDM BlockDevice should support multiple mount points
Partitions can have multiple mount points in Linux. While the BlockDevice struct in NDM [supports this](https://github.com/openebs/node-disk-manager/blob/b77aa7f29cfb9043a56b02c71c3182920d8192ed/blockdevice/blockdevice.go#L256), the BlockDevice k8s API resource can only store a single mount point.

https://github.com/openebs/node-disk-manager/blob/e4f921ce4494fc02e71a07669d4a6804de86cb9a/pkg/apis/openebs/v1alpha1/blockdevice_types.go#L202 

The API resource definition should be updated to store an array of mount points similar to the `BlockDevice` struct in package blockdevice.

Tasks:
- [x] Update mount package to return all mount points from mounts file
- [ ] Update blockdevice CR to hold a list of mount points

z0marlin:
Partitions can be mounted at multiple points in a given system.
Return a list of all the mount points of a given device as found in the
mounts file instead of just returning the first entry.

**Checklist:**
- [x] Fixes #558 
- [x] Commit has unit tests

### DIFF
--- a/pkg/mount/mountutil.go
+++ b/pkg/mount/mountutil.go
@@ -19,11 +19,12 @@ package mount
 import (
 	"bufio"
 	"fmt"
-	"github.com/openebs/node-disk-manager/pkg/features"
 	"io"
 	"os"
 	"path/filepath"
 	"strings"
+
+	"github.com/openebs/node-disk-manager/pkg/features"
 )
 
 var ErrCouldNotFindRootDevice = fmt.Errorf("could not find root device")
@@ -73,6 +74,7 @@ func (m DiskMountUtil) GetDiskPath() (string, error) {
 // getDeviceMountAttr read mounts file and returns device mount attributes, which includes partition name,
 // mountpoint and filesystem
 func (m DiskMountUtil) getDeviceMountAttr(fn getMountData) (DeviceMountAttr, error) {
+	found := false
 	mountAttr := DeviceMountAttr{}
 	// Read file from filepath and get which partition is mounted on given mount point
 	file, err := os.Open(m.filePath)
@@ -97,9 +99,13 @@ func (m DiskMountUtil) getDeviceMountAttr(fn getMountData) (DeviceMountAttr, err
 		if !strings.HasPrefix(line, "/dev") {
 			continue
 		}
-		if mountAttr, ok := fn(line); ok {
-			return mountAttr, nil
+		if lineMountAttr, ok := fn(line); ok {
+			found = true
+			mergeDeviceMountAttrs(&mountAttr, &lineMountAttr)
 		}
+	}
+	if found {
+		return mountAttr, nil
 	}
 	return mountAttr, fmt.Errorf("could not get device mount attributes, Path/MountPoint not present in mounts file")
 }
@@ -293,7 +299,7 @@ func (m *DiskMountUtil) getMountName(mountLine string) (DeviceMountAttr, bool) {
 	}
 	// mountoptions are ignored. devicepath, mountpoint and filesystem is used
 	if parts := strings.Split(mountLine, " "); parts[0] == m.devPath {
-		mountAttr.MountPoint = parts[1]
+		mountAttr.MountPoint = []string{parts[1]}
 		mountAttr.FileSystem = parts[2]
 		isValid = true
 	}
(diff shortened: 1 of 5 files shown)


Z0Marlin:
> @Z0Marlin Can you check the DCO Sign Off?

The DCO check is passing. Have I missed something?

akhilerm:
> The DCO check is passing. Have I missed something?

Nope. Its good now.

pkg/mount/mountutil.go
```diff
@@ -19,11 +19,12 @@ package mount
 import (
 	"bufio"
 	"fmt"
-	"github.com/openebs/node-disk-manager/pkg/features"
 	"io"
 	"os"
 	"path/filepath"
 	"strings"
+
+	"github.com/openebs/node-disk-manager/pkg/features"
```
akhilerm, line 27: :+1:

pkg/mount/mountutil_test.go
```diff
@@ -70,13 +70,13 @@ func TestGetMountAttr(t *testing.T) {
 	}{
 		"sda4 mounted at /": {
 			"/dev/sda4",
-			DeviceMountAttr{MountPoint: "/", FileSystem: "ext4"},
+			DeviceMountAttr{MountPoint: []string{"/"}, FileSystem: "ext4"},
```
akhilerm, line 73: Can you add a test with a single partition mounted at 2 paths.

pkg/mount/mountutil.go
```diff
@@ -97,10 +99,14 @@ func (m DiskMountUtil) getDeviceMountAttr(fn getMountData) (DeviceMountAttr, err
 		if !strings.HasPrefix(line, "/dev") {
 			continue
 		}
-		if mountAttr, ok := fn(line); ok {
-			return mountAttr, nil
+		if lineMountAttr, ok := fn(line); ok {
```
akhilerm, line 102: Here `fn` can either be either [getPartitionName](https://github.com/openebs/node-disk-manager/blob/master/pkg/mount/mountutil.go#L269) or [getMountName](https://github.com/openebs/node-disk-manager/blob/master/pkg/mount/mountutil.go#L288).  Can you test the new approach works in both cases and does not break the existing implementation

pkg/mount/mountutil.go
```diff
@@ -333,3 +339,27 @@ func fileExists(file string) bool {
...
+// performed as follows:
+// 1. If the DevPath of the first mountattr is empty, then it is set to the DevPath of
+// the second mountattr
+// 2. If the FileSystem of the first mountattr is empty, it is set to the FileSystem of
+// the second mountattr provided that the DevPaths of both the mountattrs match
+// 3. The MountPoint(s) of the second mountattr are appended to first's only if the
+// DevPath and the FileSystem of both the mountattrs match and the FileSystem of first
+// mountattr is non-empty (!= "")
+func mergeDeviceMountAttrs(ma *DeviceMountAttr, mb *DeviceMountAttr) {
+	if ma.DevPath == "" {
```
akhilerm, line 353: Why is this check needed? If `DevPath` is empty we can return empty without merging

akhilerm (commented):
given some comments.

pkg/mount/mountutil.go
Z0Marlin: This is required in cases when `getPartitionName` is used. In this case, the DevPath of `ma` will be empty but DevPath of `mb` will be set by `getPartitionName`. So we need to merge that info into the result which is `ma`.

akhilerm: ok. got it.

akhilerm: Can you add test cases when both the functions are called so that we dont break any existing functionality.

pkg/mount/mountutil_test.go
Z0Marlin: done

pkg/mount/mountutil.go
Z0Marlin: sure. I saw a to-do in the tests file for adding tests for `getPartitionName`. I'll add those tests in a separate PR and then update this one (I prefer keeping separate changes in separate PRs 🙈 )

akhilerm: Thats good.

akhilerm (approved):
LGTM

RealHarshThakur:
@Z0Marlin  Changes we discussed in the issue related to adding field in CRD, would that  be a separate PR ?

Z0Marlin:
I think yes. I'm still waiting on a confirmation to make the changes in CRD. Have broken down the issue into two tasks for the same reason. @RealHarshThakur

RealHarshThakur (approved):

akhilerm:
@Z0Marlin Can you add the changelog to this PR?

Z0Marlin:
> @Z0Marlin Can you add the changelog to this PR?

Sure. Also, I have opened #592 to add tests related to `getPartitionName`. If that can be merged first, we can verify that this PR doesn't break the existing code.

akhilerm:
@Z0Marlin Can you fix the conflicts in this PR?

akhilerm (approved):
LGTM

RealHarshThakur (approved):

PR #593, opened 2021-05-25
fix(tests): fix buggy tests in probe and sysfs

### LINKED ISSUE #589: TestUdevProbe fails with SIGSEGV in certain systems
[`TestUdevProbe`](https://github.com/openebs/node-disk-manager/blob/110066c2d6626ba26b55660b6e9a00294140b6be/cmd/ndm_daemonset/probe/udevprobe_test.go#L136) returns SIGSEGV after trying to create a GPT partition on a disk that is currently being used by the system.

Crash logs can be found in [this pastebin](https://pastebin.com/i9QNMdwG).

The issue seems to occur in cases when ndm isn't able to generate a uuid for the disk. The code for uuid generation can be found [here](https://github.com/openebs/node-disk-manager/blob/110066c2d6626ba26b55660b6e9a00294140b6be/cmd/ndm_daemonset/probe/uuid.go#L30). The failure in generating the uuid leads ndm to try creating a partition on the disk. This in turn fails since the device is busy and finally the test fails due to a nil pointer dereference on [this](https://github.com/openebs/node-disk-manager/blob/110066c2d6626ba26b55660b6e9a00294140b6be/cmd/ndm_daemonset/probe/udevprobe_test.go#L205) line.

**Environment**
- OS: Ubuntu 20.10 (Linux 5.8.0-50-generic, x86_64)
- Number of disks: 1 (with 4 partitions). **The disk does not have a wwn**.
- The machine is a VM running in VMWare Workstation.

z0marlin:
changes made:
- fix sysfs tests to cleanup after finishing. tests were failing when
run the second time after passing for the first time.
- skip `TestUdevProbe` in package probe if disk uuid cannot be
generated
- fix `TestFillDiskDetails` in package probe to correctly set
expected dev links

**Checklist:**
- [x] Fixes #589 
- [x] Commit has unit tests

### DIFF
--- a/cmd/ndm_daemonset/probe/udevprobe_test.go
+++ b/cmd/ndm_daemonset/probe/udevprobe_test.go
@@ -18,10 +18,10 @@ package probe
 
 import (
 	"errors"
-	"github.com/openebs/node-disk-manager/blockdevice"
 	"sync"
 	"testing"
 
+	"github.com/openebs/node-disk-manager/blockdevice"
 	"github.com/openebs/node-disk-manager/cmd/ndm_daemonset/controller"
 	apis "github.com/openebs/node-disk-manager/pkg/apis/openebs/v1alpha1"
 	libudevwrapper "github.com/openebs/node-disk-manager/pkg/udev"
@@ -113,14 +113,18 @@ func TestFillDiskDetails(t *testing.T) {
 	expectedDiskInfo.DeviceAttributes.WWN = mockOsDiskDetails.Wwn
 	expectedDiskInfo.PartitionInfo.PartitionTableType = mockOsDiskDetails.PartTableType
 	expectedDiskInfo.DeviceAttributes.IDType = mockOsDiskDetails.IdType
-	expectedDiskInfo.DevLinks = append(expectedDiskInfo.DevLinks, blockdevice.DevLink{
-		Kind:  libudevwrapper.BY_ID_LINK,
-		Links: mockOsDiskDetails.ByIdDevLinks,
-	})
-	expectedDiskInfo.DevLinks = append(expectedDiskInfo.DevLinks, blockdevice.DevLink{
-		Kind:  libudevwrapper.BY_PATH_LINK,
-		Links: mockOsDiskDetails.ByPathDevLinks,
-	})
+	if len(mockOsDiskDetails.ByIdDevLinks) > 0 {
+		expectedDiskInfo.DevLinks = append(expectedDiskInfo.DevLinks, blockdevice.DevLink{
+			Kind:  libudevwrapper.BY_ID_LINK,
+			Links: mockOsDiskDetails.ByIdDevLinks,
+		})
+	}
+	if len(mockOsDiskDetails.ByPathDevLinks) > 0 {
+		expectedDiskInfo.DevLinks = append(expectedDiskInfo.DevLinks, blockdevice.DevLink{
+			Kind:  libudevwrapper.BY_PATH_LINK,
+			Links: mockOsDiskDetails.ByPathDevLinks,
+		})
+	}
 
 	// The devlinks are compared separately as the ordering of devlinks can be different in some systems
 	// eg: ubuntu 20.04 in github actions
@@ -186,7 +190,7 @@ func TestUdevProbe(t *testing.T) {
 	}
 	probeEvent.addBlockDeviceEvent(eventDetails)
 	// Retrieve disk resource
-	uuid, _ := generateUUID(*deviceDetails)
+	uuid, ok := generateUUID(*deviceDetails)
 	cdr1, err1 := fakeController.GetBlockDevice(uuid)
 	fakeDr, err := mockOsDiskToAPI()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


z0marlin:
@akhilerm PTAL

pkg/sysfs/syspath_test.go
```diff
@@ -17,10 +17,12 @@ limitations under the License.
 package sysfs
 
 import (
-	"github.com/openebs/node-disk-manager/blockdevice"
-	"github.com/stretchr/testify/assert"
 	"os"
+	"path/filepath"
 	"testing"
+
+	"github.com/openebs/node-disk-manager/blockdevice"
```
akhilerm, line 24: :+1:

akhilerm: Can we follow the "standard libraries - vendor - project", pattern for ordering imports.

pkg/sysfs/syspath_test.go
```diff
@@ -76,20 +78,24 @@ func TestGetParent(t *testing.T) {
 }
 
 func TestGetDeviceSysPath(t *testing.T) {
-	sysFSDirectoryPath = "/tmp/sys/"
-
+	tmp := sysFSDirectoryPath
+	sysFSDirectoryPath = filepath.Join(t.TempDir(), "sys/") + "/"
+	t.Cleanup(func() {
+		sysFSDirectoryPath = tmp
```
akhilerm, line 84: q: How will this work?

akhilerm: When will this be executed. Also it sets the `sysFSDirectoryPath` back to the actual `/sys`

pkg/sysfs/syspath_test.go
```diff
@@ -596,7 +647,8 @@ func TestSysFsDeviceGetDeviceType(t *testing.T) {
 			sysfsDevice: &Device{
 				deviceName: "sdb1",
 				path:       "/dev/sdb1",
-				sysPath:    "/tmp/sys/devices/pci0000:00/0000:00:1f.2/ata1/host0/target0:0:0/0:0:0:0/block/sdb/sdb1/",
+				sysPath: filepath.Join(tmpDir,
+					"sys/devices/pci0000:00/0000:00:1f.2/ata1/host0/target0:0:0/0:0:0:0/block/sdb/sdb1/") + "/",
```
akhilerm, line 651:
```suggestion
					"sys/devices/pci0000:00/0000:00:1f.2/ata1/host0/target0:0:0/0:0:0:0/block/sdb/sdb1") + "/",
```

akhilerm: If using `filepath.Join()`, I dont think its necessary to mention that `/` at the end.

cmd/ndm_daemonset/probe/udevprobe_test.go
```diff
@@ -186,8 +190,12 @@ func TestUdevProbe(t *testing.T) {
 	}
 	probeEvent.addBlockDeviceEvent(eventDetails)
 	// Retrieve disk resource
-	uuid, _ := generateUUID(*deviceDetails)
+	uuid, ok := generateUUID(*deviceDetails)
+	if !ok {
+		t.Skip("cannot uniquely identify disk. skipping")
```
akhilerm, line 195: Along with skipping this test, can you add another test case, in which the uuid will be generated and the blockdevice resource will be created. If we skip this test, I think there is no test available for `addBlockDeviceEvent`

akhilerm (commented):
@z0marlin Provided some comments

akhilerm:
@z0marlin Does the changes in this PR, fix the issue in #589 ?

pkg/sysfs/syspath_test.go
z0marlin: once the test is completed. just to reset the variable back to its value. It won't make any difference for now, but just added it in case not reverting back has some side effects in the future.

pkg/sysfs/syspath_test.go
z0marlin: you mean the `+ "/"` part?

cmd/ndm_daemonset/probe/udevprobe_test.go
z0marlin: Sure. I'll have to see what test I can add and how I can make it work.

z0marlin:
> @z0marlin Does the changes in this PR, fix the issue in #589 ?

Yes it does. Since the main issue was disk not being added because uuid couldn't be generated, we are skipping the test. Only in cases when the uuid isn't available, the pointer becomes `nil`

pkg/sysfs/syspath_test.go
z0marlin: My bad. will reorder

z0marlin:
On further thought, I think we can also add nil-pointer checks and fail the test in such cases. These checks can be added after the uuid check so that we can be sure that a uuid was generated.

pkg/sysfs/syspath_test.go
akhilerm: nope. The `/` at the end of `block/sdb/sdb1/`

akhilerm (approved):
LGTM

akhilerm:
@z0marlin Will the additional test for the skipped check be taken up in a different PR?

z0marlin:
Alright, so If u see how I've rearranged the code, I think no additional test is needed. We aren't skipping the test. In case uuid is not generated, we expect the disk to not be added since the disk being used during tests is always the root disk. Hence, we just verify that the pointer is indeed nil. This according to me is a part of the test itself. So no additional test needed. Wdyt?

akhilerm:
> Alright, so if you see how I've rearranged the code, I think no additional test is needed. We aren't skipping the test. In case uuid is not generated, we expect the disk to not be added since the disk being used during tests is always the root disk. Hence, we just verify that the pointer is indeed nil. This according to me is a part of the test itself. So no additional test needed. Wdyt?

Then its good. :+1:

RealHarshThakur (approved):

PR #594, opened 2021-05-31
docs: add mount change detection design

z0marlin:
**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/docs/mount-change-detection.md
+++ b/docs/mount-change-detection.md
@@ -32,7 +32,7 @@ A new package `libmount` is introduced for parsing the mounts file. The package 
 has additional information regarding what probes are to be run and specifies that only _mount-probe_ needs to be run for the event. This is done since _mount-probe_ alone can fetch the new mounts and fs data for the blockdevices. Running the probes selectively helps us optimize the updation process.
 The message is then received by the loop in `udevProbe.listen()` and sent further down to the `ProbeEvent` change handler.
 
-For every blockdevice listed in the `EventMessage`, the change handler first fetches the latest copy of the blockdevice from the controller blockdevice cache (`controller.BDHeirarchyCache`) and then runs the it though the requested probes which are also provided in the message. Once the blockdevice is run through all the probes, the cache is updated and an update request is send to the kuebrnetes api server to upate the corresponding blocdevice CR.
+For every blockdevice listed in the `EventMessage`, the change handler first fetches the latest copy of the blockdevice from the controller blockdevice cache (`controller.BDHierarchyCache`) and then runs the it though the requested probes which are also provided in the message. Once the blockdevice is run through all the probes, the cache is updated and an update request is send to the kuebrnetes api server to upate the corresponding blocdevice CR.
 
 &nbsp;
 
--- a/docs/mount-change-detection.md
+++ b/docs/mount-change-detection.md
@@ -28,33 +28,67 @@ rootfs / rootfs rw 0 0
 
 Whenever a block device is (un)mounted or the fs type changes, the changes are reflected in the mounts file. This proposal introduces a change to the existing _**mount-probe**_ in NDM. Similar to how _udev-probe_ listens to udev events by starting a loop in its `Start()`, a loop is started by _mount-probe_ that watches for changes in the mounts file and triggers updation when a change is detected. The *epoll* API is used to watch the mounts file for changes. The *epoll* API is provided by the Linux kernel for the userspace programs to monitor file descriptors and get notifications about I/O events happening on them. Whenever the mounts file changes, the events `EPOLLPRI` and `EPOLLERR` are emitted. This behaviour has been documented [here](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=31b07093c44a7a442394d44423e21d783f5523b8) (additional links - [\[1\]](https://lkml.org/lkml/2006/2/22/169), [\[2\]](http://lkml.iu.edu/hypermail/linux/kernel/1012.1/02246.html)).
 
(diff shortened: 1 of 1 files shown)


docs/mount-change-detection.md
```diff
@@ -0,0 +1,58 @@
...
+## Introduction
+
+This document presents the design of the mount-point change detection system in NDM.
+The goal of mount-point change detection is to detect the changes in the mount-points
+and the filesystem on the existing blockdevices discovered by NDM and trigger appropriate
+action to update the blockdevice CRs.
+
+## Design
+
+The mount-points and filesystem info of a block device are found by the probe - *mountprobe*.
```
RealHarshThakur, line 12: I checked the `findmnt` manual page and it says it looks for mount points in `/etc/fstab` ,  `/etc/mtab` and  `/proc/self/mountinfo` . Just wondering if there's a reason NDM should look at fstab or mtab too.

docs/mount-change-detection.md
```diff
@@ -0,0 +1,58 @@
...
+/proc /proc proc rw 0 0 usbdevfs 
+/proc/bus/usb usbdevfs rw 0 0 
+/dev/sda1 /boot ext3 rw 0 0 none 
+/dev/pts devpts rw 0 0 
+/dev/sda4 /home ext3 rw 0 0 none 
+/dev/shm tmpfs rw 0 0 none 
+/proc/sys/fs/binfmt_misc binfmt_misc rw 0 0
+```
+
+Whenever a block device is (un)mounted or the fs type changes, the changes are reflected in the mounts file. This proposal introduces a new probe in NDM called _**mount-change-probe (MCP)**_. The probe will watch for changes in the mounts file and trigger updation when a change is detected. To watch the mounts file for changes, MCP uses the *epoll* API. The *epoll* API is provided by the Linux kernel for the userspace programs to monitor file descriptors and get notifications about I/O events happening on them. Whenever the mounts file changes, the events `EPOLLPRI` and `EPOLLERR` are emitted. This behaviour has been documented [here](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=31b07093c44a7a442394d44423e21d783f5523b8) (additional links - [\[1\]](https://lkml.org/lkml/2006/2/22/169), [\[2\]](http://lkml.iu.edu/hypermail/linux/kernel/1012.1/02246.html)).
```
akhilerm, line 29: When the fs type changes, there is no change in the mounts file. `mounts` file reflect only about changes of devices that are mounted.

You can use the [`Start()`](https://github.com/openebs/node-disk-manager/blob/master/cmd/ndm_daemonset/probe/mountprobe.go#L89) in mountprobe itself right? instead of a new probe? Similar way how udevprobe does monitoring?

akhilerm (commented):
Provide a note in the design:
 filesystem changes can be detected only once the device is mounted. If just format a filesystem on the device and does not mount it, it wont be updated.

@z0marlin Have one comment on using the existing mountprobe itself. everything else looks good.  Can proceed with implementation

docs/mount-change-detection.md
z0marlin: > When the fs type changes, there is no change in the mounts file. `mounts` file reflect only about changes of devices that are mounted.

👍🏽 will add note

> You can use the [`Start()`](https://github.com/openebs/node-disk-manager/blob/master/cmd/ndm_daemonset/probe/mountprobe.go#L89) in mountprobe itself right? instead of a new probe? Similar way how udevprobe does monitoring?

Ture. This looks like the cleaner way to do it.

docs/mount-change-detection.md
z0marlin: Not sure if we should be watching fstab/mtab. I took a look at `findmnt` though. It provies an option to watch for changes (`-p`). To implement this, it also [uses the linux polling api](https://github.com/karelzak/util-linux/blob/059ba3dd76c9d6cf45ce01e01bb03091a14a48a9/misc-utils/findmnt.c#L1145) and by default wacthes for changed in `/proc/self/mountinfo`.

RealHarshThakur: Cool, that's good validation. We can add watching fstab/mtab if the need ever arises I suppose. Right now, don't see a reason we need it

RealHarshThakur (dismissed):

akhilerm (approved):
LGTM

RealHarshThakur (dismissed):

docs/mount-change-detection.md
```diff
@@ -0,0 +1,94 @@
...
+has additional information regarding what probes are to be run and specifies that only _mount-probe_ needs to be run for the event. This is done since _mount-probe_ alone can fetch the new mounts and fs data for the blockdevices. Running the probes selectively helps us optimize the updation process.
+The message is then received by the loop in `udevProbe.listen()` and sent further down to the `ProbeEvent` change handler.
+
+For every blockdevice listed in the `EventMessage`, the change handler first fetches the latest copy of the blockdevice from the controller blockdevice cache (`controller.BDHeirarchyCache`) and then runs the it though the requested probes which are also provided in the message. Once the blockdevice is run through all the probes, the cache is updated and an update request is send to the kuebrnetes api server to upate the corresponding blocdevice CR.
```
akhilerm, line 35:
```suggestion
For every blockdevice listed in the `EventMessage`, the change handler first fetches the latest copy of the blockdevice from the controller blockdevice cache (`controller.BDHierarchyCache`) and then runs the it though the requested probes which are also provided in the message. Once the blockdevice is run through all the probes, the cache is updated and an update request is send to the kuebrnetes api server to upate the corresponding blocdevice CR.
```
typo

akhilerm (dismissed):
LGTM.

@z0marlin Please fix the typo.

akhilerm (approved):
LGTM

PR #601, opened 2021-06-24
feat(controller): selectively run probes using allowlist in event msg

z0marlin:
event messages passed in the `UdevEventMessageChannel` channel are used
to trigger actions in the controller to add/update/delete blockdevices.
In case of addition, the controller runs all the registered probes to
fill certain details of the blockdevice. While all probes need to be run
when adding a new blockdevice, in case of updates, the list of probes to
be run can be narrowed down based on the type of update. For example,
when the mount points of a partition change, mount probe alone is enough
to fetch the new information.
add an allowlist for probes in the event message to inform the
controller what probes are to be run. the controller then runs only the
probes that are enabled and available in the allowlist. leaving the
allowlist empty causes default behaviour which is to run all the enabled
probes.

This PR is a continuation of #595 and works towards fixing openebs/openebs#3135.

**Checklist:**
- [x] Fixes  openebs/openebs#3135 (partially)
- [x] Commit has unit tests

### DIFF
--- a/cmd/ndm_daemonset/controller/probe.go
+++ b/cmd/ndm_daemonset/controller/probe.go
@@ -27,9 +27,9 @@ import (
 
 // EventMessage struct contains attribute of event message info.
 type EventMessage struct {
-	Action        string                     // Action is event action like attach/detach
-	Devices       []*blockdevice.BlockDevice // list of block device details
-	AllowedProbes []string                   // List of probes (given as probe names) to be run for this event. Optional
+	Action          string                     // Action is event action like attach/detach
+	Devices         []*blockdevice.BlockDevice // list of block device details
+	RequestedProbes []string                   // List of probes (given as probe names) to be run for this event. Optional
 }
 
 // Probe contains name, state and probeinterface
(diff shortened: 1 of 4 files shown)


cmd/ndm_daemonset/controller/probe.go
```diff
@@ -27,8 +27,9 @@ import (
 
 // EventMessage struct contains attribute of event message info.
 type EventMessage struct {
-	Action  string                     // Action is event action like attach/detach
-	Devices []*blockdevice.BlockDevice // list of block device details
+	Action        string                     // Action is event action like attach/detach
+	Devices       []*blockdevice.BlockDevice // list of block device details
+	AllowedProbes []string                   // List of probes (given as probe names) to be run for this event. Optional
```
akhilerm, line 32: Refer to the probe names in `probe` package. Also we may need to export the probe names as they are currently unexported. [Ref](https://github.com/openebs/node-disk-manager/blob/master/cmd/ndm_daemonset/probe/mountprobe.go#L42)

cmd/ndm_daemonset/controller/probe.go
```diff
@@ -85,24 +86,35 @@ func (c *Controller) AddNewProbe(probe *Probe) {
...
 	for _, probe := range c.Probes {
 		if probe.State {
-			listProbe = append(listProbe, probe)
+			if _, ok := requestedProbes[probe.Name]; allProbes || ok {
+				listProbe = append(listProbe, probe)
+			}
 		}
 	}
 	return listProbe
 }
```
akhilerm, line 111: The logic can be made much simpler by writing like this.
```golang
func (c *Controller) ListProbe(requestedProbes ...string) []*Probe {
	c.Lock()
	defer c.Unlock()
	allProbes := false
	if len(requestedProbes) == 0 {
		allProbes = true
	}
	listProbe := make([]*Probe, 0)
	for _, probe := range c.Probes {
		if probe.State {
			if allProbes || util.Contains(requestedProbes, probe.Name) {
				listProbe = append(listProbe, probe)
			}
		}
	}
	return listProbe
}
```

cmd/ndm_daemonset/controller/probe.go
```diff
@@ -85,24 +86,35 @@ func (c *Controller) AddNewProbe(probe *Probe) {
...
 
 // FillBlockDeviceDetails lists registered probes and fills details from each probe
-func (c *Controller) FillBlockDeviceDetails(blockDevice *blockdevice.BlockDevice) {
+func (c *Controller) FillBlockDeviceDetails(blockDevice *blockdevice.BlockDevice, probes ...string) {
```
akhilerm, line 114: How does the variable name `requestedProbes` sound to you?

cmd/ndm_daemonset/probe/eventhandler.go
```diff
@@ -58,7 +58,7 @@ func (pe *ProbeEvent) addBlockDeviceEvent(msg controller.EventMessage) {
 	// iterate through each block device and perform the add/update operation
 	for _, device := range msg.Devices {
 		klog.Infof("Processing details for %s", device.DevPath)
-		pe.Controller.FillBlockDeviceDetails(device)
+		pe.Controller.FillBlockDeviceDetails(device, msg.AllowedProbes...)
```
akhilerm, line 61: Same change w.r.t variable name can be made here. RequestedProbes instead of AllowedProbes.

akhilerm (changes requested):
@z0marlin Given some comments and need a few changes

cmd/ndm_daemonset/controller/probe.go
z0marlin: Do we need to export them? The only other place outside the probe package where events are pushed to the event channel is the udevevent package. most of the usage will be internal to the probe package I feel. Wdyt?

akhilerm (approved):
Changes looks good, shall we merge this after #595

z0marlin:
> Changes looks good, shall we merge this after #595

anything works. can be merged before it also. and I'll update the other PR. actually it'll be better if we merge this first. I feel that some of the issues being caused in #595 are due to the fact that we are running all probes when we detect a change which in turn is causing a delay and not letting us get the next event. though the chances of this happening are low, we can try and see if running only a single probe helps.

cmd/ndm_daemonset/controller/probe.go
akhilerm: yep. no need to export. I got it clear now.

Issue #597, opened 2021-06-16
[Design] use NDM packages in Device CSI Driver


kmova:

akhilerm:
The `udevevent` package has already been decoupled from the probe and controller. 

Identified list of remaining changes:
1. The other packages that need to be refactored, that are currently under `cmd/ndm-daemonset`.
2. Check in device-localpv for functionality that can be moved into NDM (eg: creating partitions, finding free slots).

A few more refactoring changes are needed that are listed [here](https://github.com/orgs/openebs/projects/2)

PR #609, opened 2021-07-22
refactor(pkg, udevevent): decouple udevevent from controller

z0marlin:
As per conventions, packages in the pkg directory should be independent
of the overlying applications that use them. The `udevevent` package has
tight coupling with the `controller` and `probe` packages from
ndm_daemonset.
Refactor to make `udevevent` standalone. This includes
moving all application (ndm_daemonset in this case) specific code in
`udevevent` to appropriate cmd packages and providing an API for the
application code to use the functionalities of `udevevent`.

This PR aligns with the idea raised in #597 

**Checklist:**
- [x] Related to #597 
- [x] Commit has unit tests


cmd/ndm_daemonset/controller/probe.go
```diff
@@ -19,10 +19,10 @@ package controller
 import (
 	"sort"
 
+	"k8s.io/klog"
```
z0marlin, line 22: reordering as per convention

pkg/udevevent/monitor.go
```diff
@@ -20,21 +20,40 @@ import (
 	"errors"
 	"syscall"
 
-	"github.com/openebs/node-disk-manager/cmd/ndm_daemonset/controller"
+	"k8s.io/klog"
+
```
RealHarshThakur, line 24: If we're going to use this package as a library, I am not sure if we should be logging. It makes sense to me because the cmd/ packages can use these libraries, receive an error and log them. Would like to know your thoughts too @kmova @akhilerm

z0marlin: Agree with you here.

cmd/ndm_daemonset/probe/udevprobe.go
```diff
@@ -360,3 +364,79 @@ func (up *udevProbe) free() {
 		up.udevEnumerate.UnrefUdevEnumerate()
 	}
 }
+
+func (up *udevProbe) listenUdevEventMonitor() {
+	for {
+		event := <-up.udeveventSubscription.Events()
```
akhilerm, line 370: Do we need to get an instance of the target channel on every iteration?

cmd/ndm_daemonset/probe/udevprobe.go
```diff
@@ -360,3 +364,79 @@ func (up *udevProbe) free() {
...
+
+func processUdevEvent(event udevevent.UdevEvent) controller.EventMessage {
+	defer event.UdevDeviceUnref()
+	diskInfo := make([]*blockdevice.BlockDevice, 0)
+	uuid := event.GetUid()
+	path := event.GetPath()
+	action := event.GetAction()
+	klog.Infof("processing new event for (%s) action type %s", path, action)
+	deviceDetails := &blockdevice.BlockDevice{}
+	deviceDetails.UUID = uuid
```
akhilerm, line 383: provide a comment here. This uuid will be the legacy format. We will be overwriting this UUID in the event handler.

pkg/udevevent/monitor.go
```diff
@@ -20,21 +20,40 @@ import (
...
+type UdevEvent struct {
+	*libudevwrapper.UdevDevice
+	eventType UdevEventType
+}
+
+type Subscription struct {
+	targetChannel   chan UdevEvent
+	subscribedTypes []UdevEventType
+}
+
```
akhilerm, line 46: add comments

pkg/udevevent/monitor.go
```diff
@@ -98,12 +117,35 @@ func (m *monitor) process(fd int) error {
...
-	event.process(device)
-	event.send()
+	var eventType UdevEventType
+	switch device.GetAction() {
+	case libudevwrapper.UDEV_ACTION_ADD:
+		eventType = EventTypeAdd
+	case libudevwrapper.UDEV_ACTION_REMOVE:
+		eventType = EventTypeRemove
+	default:
+		eventType = EventTypeChange
```
akhilerm, line 127: Should we default to change, or just ignore?

pkg/udevevent/monitor.go
```diff
@@ -98,12 +117,35 @@ func (m *monitor) process(fd int) error {
...
+	}
+	event := UdevEvent{device, eventType}
+	dispatchEvent(event)
 	return nil
 }
 
+func dispatchEvent(event UdevEvent) {
+	for _, sub := range subscriptions {
```
akhilerm, line 135: Is this created so that there can be multiple subscribers for the events?

pkg/udevevent/monitor.go
```diff
@@ -122,3 +164,61 @@ func Monitor() {
 		}
 	}
 }
+
+func Subscribe(eventTypes ...UdevEventType) *Subscription {
```
akhilerm, line 168: Add comments

akhilerm: Why do we need to return the Subscription here?

pkg/udevevent/monitor.go
```diff
@@ -122,3 +164,61 @@ func Monitor() {
...
+	}
+	subscription := Subscription{
+		targetChannel:   make(chan UdevEvent),
+		subscribedTypes: eventTypes,
+	}
+	subscriptions = append(subscriptions, &subscription)
+	return &subscription
+}
+
+func Unsubscribe(sub *Subscription) error {
```
akhilerm, line 180: do we really use the Unsubscribe method?

pkg/udevevent/monitor.go
```diff
@@ -122,3 +164,61 @@ func Monitor() {
...
+}
+
+func (u UdevEvent) GetAction() UdevEventType {
+	return u.eventType
+}
+
+func (uevt UdevEventType) String() string {
+	switch uevt {
+	case EventTypeAdd:
+		return "add"
```
akhilerm, line 216: Cant we use the `libudevwrapper.UDEV_ACTION_*` here?

pkg/udevevent/monitor.go
z0marlin: yes

pkg/udevevent/monitor.go
z0marlin: we need the subscription in the `Unsubscribe` call

pkg/udevevent/monitor.go
z0marlin: currently, we probably won't. but kept it to have a complete API.

cmd/ndm_daemonset/probe/udevprobe.go
z0marlin: fixed

pkg/udevevent/monitor.go
z0marlin: throwing error in case it doesn't match

RealHarshThakur (approved):

pkg/udevevent/monitor.go
```diff
@@ -147,22 +145,26 @@ func dispatchEvent(event UdevEvent) {
...
-	defer monitor.free()
-	fd, err := monitor.setup()
-	if err != nil {
-		klog.Error(err)
-	}
-	for {
-		err := monitor.process(fd)
+func Monitor() <-chan error {
+	errChan := make(chan error)
+	go func() {
```
akhilerm, line 154: Is this go routine created for processing errors only?

akhilerm: Didnt understand this function @z0marlin. Too many channels now passing information.

Also, how many go routines (created from the code) will be present in NDM daemon now?

akhilerm (commented):
have a few questions on the new goroutine and channel for passing error messages.

pkg/udevevent/monitor.go
z0marlin: The number of go routines hasn't changed. Its just that instead of starting the monitor as a go routine in udevprobe [here](https://github.com/openebs/node-disk-manager/pull/609/files/c7f7cab6a7f0a24c874b63d2f2ee684e3cc8e606#diff-7f7a5459cf891826a55f62700489f89dcd306dac941bf39d68c9091725b7456eL121), I'm starting it within the call to `Monitor()`. This allows me to return an error channel back to udevprobe.

akhilerm (approved):
lgtm

PR #611, opened 2021-08-03
docs: add design docs for device size change detection

z0marlin:
**Checklist:**
- [ ] Related issues - openebs/openebs#3135


docs/device-size-change-detection.md
```diff
@@ -0,0 +1,32 @@
+# Blockdevice size change detection
+
+## Introduction
+
+This document presents the design for detecting changes in the capacity (or size) of
+blocdevices in NDM.
```
akhilerm, line 6:
```suggestion
blockdevices in NDM.
```

docs/device-size-change-detection.md
```diff
@@ -0,0 +1,32 @@
...
+
+To capture change events from *udev*, the *sysfs* probe subscribes to change events using the existing *udevevent* package. On receiving a udev event, the sysfs probe processes the event to get the list of devices that were modified and creates a corresponding `controller.EventMessage` with the `requestedProbes` field containing only *sysfs* probe since only the size needs to be updated. This message is then pushed into `controller.EventMessageChannel` and further processed by the event handler.
```
akhilerm, line 24: Are we planning to put any filtering mechanism to check if the device/size was changed,? and then send the event message.

akhilerm: Also, can we have multiple subscribers to the udev event package? 

If two probes subscribe to udev event we need to make sure, if at some point, udev probe also decides to subscribe, it should exclude the sysfs probe, other wise it will result in same probe being run twice for the same data.

docs/device-size-change-detection.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Alternative Design
+
+A major issue with depending on *udev* for detecting size changes is that a good number of events won't be related to changes in the size. An alternative approach to detect changes to the size is to periodically read the `size` attribute file in *sysfs* and check if it has changed (polling). The polling interval can be fixed or configurable by user. A disadvantage of this approach is that polling will need to be done for the size file of each and every blockdevice that has been detected and depending on the poll interval, this operation can be expensive.
```
akhilerm, line 28: Can we set polling to listen only on files for which blockdevices were actually created instead of the detected blockdevices on the system.

eg: `/dev/sda` is present on the system where `/` is installed. But it is filtered out and blockdevice is not created for this disk. In this case, we do not need to listen for `size` file changes on `sda`.

akhilerm (commented):
Given a few comments

docs/device-size-change-detection.md
z0marlin: this can be done

docs/device-size-change-detection.md
z0marlin: yes, it is possible to have multiple subscribers to the udev events. But the subscription can also filter out events based on the type (add, remove, change). I planned to let the udevprobe only listen to add/remove events and the sysfs probe only to change events. This would prevent any overlap.

akhilerm: sure. that sounds good.

akhilerm (approved):
LGTM

PR #652, opened 2021-10-08
feat(pkg/features): added support for feature gate dependencies

### LINKED ISSUE #651: Enhance feature gate package to add support for dependent feature gates
**Describe the problem/challenge you have**
The current [feature](https://github.com/openebs/node-disk-manager/tree/develop/pkg/features) supports adding features, and setting the default value of features. But there can be cases like one feature can be dependent on another. This functionality should also be included. 

for example:
if A is dependent on B, and B is set to False, then A should default to the default value.  

**Describe the solution you'd like**
A new `dependentFeature` variable should be made of type `map[Feature][]Feature`, so each feature can be dependent on multiple features. Then while [parsing](https://github.com/openebs/node-disk-manager/blob/develop/pkg/features/features.go#L126) the features, the new checks should be performed and if dependencies are not met, the feature should fall back to the default value.

**UseCase:**
A new feature `UsePartTableUUID` is being introduced which is dependent on `GPTBasedUUID`. So the package should make sure that `UsePartTableUUID` can be enabled only if `GPTBasedUUID` is also enabled.

jdkramhoft:
**What this PR does?**:
Features can now be added that are dependent on other features. 
Attempting to set a feature flag to true if the dependency is not met will result in an error.

No, I don't think so.

The unit tests included cover the scenarios:
- Setting a feature flag to true when a dependency is met, and when it is unmet.
- Setting a feature flag to true when only one of two dependencies are met.
- Having multiple feature flags with the same dependencies

This supports dependent feature gates for later when UsePartTableUUID will be introduced.

**Checklist:**
- [x] Fixes #651 <issue number>
- [x] Commit has unit tests


pkg/features/features.go
```diff
@@ -121,6 +123,18 @@ func (fg featureFlag) SetFeatureFlag(features []string) error {
...
 		}
+		// check if the feature has dependencies
+		dependencies, hasDependencies := featureDependencies[f]
+		// if the feature is being set to true, we need to ensure dependencies are met
+		if hasDependencies && isEnabled {
+			for _, dependency := range dependencies {
+				missingDependency := !fg[dependency]
+				if missingDependency {
+					fg[f] = false
+					return fmt.Errorf("Feature %s has unmet dependency %s - enable that first", feature, dependency)
```
z0marlin: returning an error here would cause ndm to exit with failure. Is that what we want? Or do we want to just log, disable the flag and continue?

pkg/features/features.go
```diff
@@ -121,6 +123,18 @@ func (fg featureFlag) SetFeatureFlag(features []string) error {
...
+				}
+			}
+		}
```
z0marlin: This entire block should be out of the for loop. This is because the features aren't ordered. It is possible that a dependency for feature A appears in the slice after the feature A. Also, you might have to think of some recursive logic to resolve the dependencies if we want to support a case like A -> B -> C

z0marlin (commented):
Thanks for the PR @jdkramhoft :) Left a few comments.

pkg/features/features.go
jdkramhoft: Oh! Good catch; Right now if someone later disabled a necessary dependency I wouldn't catch it.
I will move it to after the loop and validate the state then, logging and disabling flags.

Seems like I technically should check for cycles and then do a topological sort. Feels a bit heavyweight for only a few support flags. Can I assume there are no cycles of dependencies A -> B -> C -> A ?

akhilerm: yes, @jdkramhoft . cycles wont be there as we itself add the feature gates in the code. What we want to achieve is the user should not be enabling some features while dependent features are disabled.

pkg/features/features.go
jdkramhoft: I have changed behavior to logging, disabling the flags and continuing.

pkg/features/features.go
```diff
@@ -123,24 +123,35 @@ func (fg featureFlag) SetFeatureFlag(features []string) error {
...
+	dependencies := featureDependencies[feature]
+	for _, dependency := range dependencies {
+		missingDependency := !ValidateDependencies(dependency, flags)
+		if missingDependency {
+			flags[feature] = false
+			klog.Infof("Feature %v was set to false due to missing dependency %v", feature, dependency)
+		}
+	}
+	return flags[feature]
+}
```
z0marlin: This looks good. one thing you could do is to use another map for storing the computed values of the validated dependencies. This way you can store the results and save computation in case the `ValidateDependencies` function is called again for the same feature flag.

jdkramhoft: I have added a map for memoized values.

jdkramhoft:
Should be working nicely now - I can't see what Better Code Hub wants.

pkg/features/features.go
```diff
@@ -124,9 +126,37 @@ func (fg featureFlag) SetFeatureFlag(features []string) error {
...
 	for k, v := range fg {
 		klog.Infof("Feature gate: %s, state: %s", k, util.StateStatus(v))
 	}
 
 	return nil
 }
+
+// Ensures features are disabled if their dependencies are unmet
+// Returns true if a feature is enabled after validation
+func ValidateDependencies(feature Feature, flags featureFlag, memoizedValues featureFlag) bool {
```
akhilerm, line 145: @z0marlin Can you take a look at this function.

pkg/features/features.go
```diff
@@ -124,9 +126,37 @@ func (fg featureFlag) SetFeatureFlag(features []string) error {
 		fg[f] = isEnabled
 	}
 
+	// We make sure features are only turned on if their dependencies are met
+	// We memoize values to avoid computing the same dependency twice
+	memoizedValues := make(featureFlag)
+	for feature := range fg {
+		ValidateDependencies(feature, fg, memoizedValues)
```
akhilerm:
```suggestion
		_ = ValidateDependencies(feature, fg, memoizedValues)
```
to ignore the return value of the function.

pkg/features/features.go
```diff
@@ -124,9 +126,37 @@ func (fg featureFlag) SetFeatureFlag(features []string) error {
...
+	}
+	if disabled := !flags[feature]; disabled {
+		return false
+	}
+	dependencies := featureDependencies[feature]
+	for _, dependency := range dependencies {
+		missingDependency := !ValidateDependencies(dependency, flags, memoizedValues)
+		if missingDependency {
+			flags[feature] = false
+			klog.Infof("Feature %v was set to false due to missing dependency %v", feature, dependency)
```
akhilerm, line 157: Can we break here, other wise we may set it to false multiple times for every unmet dependency

akhilerm (commented):
given a few comments.

pkg/features/features.go
z0marlin: Looks good

z0marlin (approved):
LGTM. Thanks for working on this @jdkramhoft :)

akhilerm (approved):
LGTM

PR #662, opened 2021-12-23
fix(cleaner): fix disk partition cannot be released after creating lvm 

liuminjian:
## Pull Request template

I create a virtual machine to mount the disk, partition the disk, then use the partition to create vg, create lv, format lv, and then mount lv to the directory. When I delete the virtual machine and release the disk, the cleaner job will report an error device busy. The operation is as follows.

1. create a vm with a block device /dev/vdb
2. create a disk partiton /dev/vdb1
3. vgcreate vg1 /dev/vdb1
4. lvcreate -L 5G -n lv1 vg1
5. mkfs.ext4 /dev/vg1/lv1
6. mount /dev/vg1/lv1 /root/test
7. delete vm and release block device

QQ20211222-190930@2x

need to update linux-utils add lvm2 and device-mapper
https://github.com/liuminjian/linux-utils/commit/90a4942cd417e168e015594a51190514e2843924

### DIFF
--- a/pkg/cleaner/jobcontroller.go
+++ b/pkg/cleaner/jobcontroller.go
@@ -104,7 +104,12 @@ func NewCleanupJob(bd *v1alpha1.BlockDevice, volMode VolumeMode, tolerations []v
 		// wipefs erases the filesystem signature from the block
 		// -a    wipe all magic strings
 		// -f    force erasure
-		args := fmt.Sprintf("(fdisk -o Device -l %[1]s "+
+
+		args := fmt.Sprintf(""+
+			"(pvs -o pv_name,vg_name|grep %[1]s|awk '{print $2}'"+
+			"| xargs -I {} sh -c 'dmsetup info -c -o name --noheadings|grep ^{}- "+
+			"| xargs -t -I {} dmsetup remove {} ') && "+
+			"(fdisk -o Device -l %[1]s "+
 			"| grep \"^%[1]s\" "+
 			"| xargs -I '{}' wipefs -fa '{}') "+
 			"&& wipefs -fa %[1]s ",


akhilerm:
@liuminjian Can you raise the corresponding PR in linux-utils repository also to update with device-mapper packages?

liuminjian:
> @liuminjian Can you raise the corresponding PR in linux-utils repository also to update with device-mapper packages?
ok https://github.com/openebs/linux-utils/pull/22

RealHarshThakur (approved):

pkg/cleaner/jobcontroller.go
```diff
@@ -104,7 +104,12 @@ func NewCleanupJob(bd *v1alpha1.BlockDevice, volMode VolumeMode, tolerations []v
 		// wipefs erases the filesystem signature from the block
 		// -a    wipe all magic strings
 		// -f    force erasure
-		args := fmt.Sprintf("(fdisk -o Device -l %[1]s "+
+
+		args := fmt.Sprintf(""+
+			"(pvs -o pv_name,vg_name|grep %[1]s|awk '{print $2}'"+
+			"| xargs -I {} sh -c 'dmsetup info -c -o name --noheadings|grep ^{}- "+
+			"| xargs -t -I {} dmsetup remove {} ') && "+
```
akhilerm, line 111: Do we need to remove only the logical volumes? What about the volume groups.? Doesn't it cause resource busy error?

akhilerm (commented):
q: have a doubt on removing volume groups from the device.

pkg/cleaner/jobcontroller.go
liuminjian: I have tested just need to delete the logical volumes. wipefs disk seemed looks to clear the volume groups.

akhilerm (approved):
LGTM

PR #682, opened 2022-12-10
chore(dependencies): Update go/k8s dependencies and ci

niladrih:
- update manusa/actions-setup-minikube to v2.7.1
- update minikube version to v1.26.1
- update kubernetes version to v1.25.0
- update minikube setup to use cilium cni
- update minikube setup to avoid storage provisioner plugin installation
- update k8s libraries to v0.25.4 (k8s v1.25.4)
- remove the vendor directory
- update go modules -- sigs.k8s.io/controller-runtime, k8s.io/klog

Signed-off-by: Niladri Halder <niladri.halder26@gmail.com>


.github/workflows/build.yml
```diff
@@ -347,10 +347,12 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Setup Minikube-Kubernetes
-        uses: manusa/actions-setup-minikube@v2.3.0
+        uses: manusa/actions-setup-minikube@v2.7.1
         with:
-          minikube version: v1.15.0
-          kubernetes version: 1.18.0
+          minikube version: 'v1.26.1'
+          kubernetes version: 'v1.25.0'
```
akhilerm: Along with this should we update the k8s dependecies in our code. Currently its at k8s 1.20. The controller runtime also would have to be updated to a matching version for the k8s version.

akhilerm (commented):
Can you check why the integration tests are failing. May be due to mismatching k8s version. But not sure.

go.mod
```diff
@@ -25,3 +25,31 @@ require (
...
+
+replace k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.25.4
+
+replace k8s.io/apimachinery => k8s.io/apimachinery v0.25.5-rc.0
+
+replace k8s.io/apiserver => k8s.io/apiserver v0.25.4
+
+replace k8s.io/cli-runtime => k8s.io/cli-runtime v0.25.4
+
+replace k8s.io/client-go => k8s.io/client-go v0.25.4
```
akhilerm: Instead of using replae directive here, this should be added in the require itself

.github/workflows/build.yml
```diff
@@ -54,6 +54,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
+      - name: Set up Go v1.19
```
akhilerm, line 57: How was this working earlier without setting up go?

.github/workflows/pull_request.yml
```diff
@@ -53,6 +53,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
+      - name: Set up Go v1.19
+        uses: actions/setup-go@v3
+        with:
+          go-version: '1.19.3'
+
```
akhilerm, line 60: Same question on how was this working earlier.

pkg/smart/doc.go
```diff
@@ -55,29 +55,30 @@ An example usage can be like this -
...
 	"github.com/openebs/node-disk-manager/pkg/smart"
+
 )
 
-func main() {
-	deviceBasicSCSIInfo, err := smart.SCSIBasicDiskInfo("/dev/sda")
-	if err != nil {
-		klog.Fatal(err)
-	}
+	func main() {
```
akhilerm: Why is this formatting needed?

akhilerm (commented):
Can you check the formatting on the .go files once. Seems like its off.

.github/workflows/build.yml
niladrih: It was using the bundled Go version. Deliberately setting up the Go version, keeps things predictable.
https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#go

pkg/smart/doc.go
niladrih: It's the IDE. I'll fix this

niladrih: I think I got all of it. Comment formatting reverted.

.github/workflows/pull_request.yml
niladrih: Refer comment reply above.

.github/workflows/build.yml
niladrih: Updated the libraries.

go.mod
niladrih: ✔️

akhilerm (approved):
LGTM

